PR #2, opened 2015-06-16
Cleanup the font awesome assets.

damonkelley:
Removes all of the additional files provided by Font Awesome. Coda only requires the fonts and css files.


damonkelley:
ping @ldko @vphill

ldko:
:+1:

PR #12, opened 2015-07-01
Fix coda_mdstore Tests

### LINKED ISSUE #11: Fix existing coda_mdstore Tests
Fix the existing tests.

damonkelley:
Closes #11 

Most of the tests were failing because the views were unable to grab the `HTTP_HOST` header from the request object. That was resolved by adding that kwarg to all of the client HTTP calls.

### DIFF
--- a/coda/coda_mdstore/tests.py
+++ b/coda/coda_mdstore/tests.py
@@ -306,9 +306,3 @@ class APPNodeHTTPTest(CodaCase):
         response = self.c.delete('/APP/node/coda-123/', HTTP_HOST='example.com')
         self.assertEqual(response.status_code, 200)
 
-# class TestSitemap(TestCase):
-#     def sitemap(self):
-#         self.assertEqual(BaseSitemap.items, 400000)
-
-if __name__ == "__main__":
-    unittest.main()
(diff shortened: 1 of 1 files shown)


damonkelley:
ping @somexpert @ldko

coda/coda_mdstore/tests.py
```diff
@@ -1,14 +1,15 @@
-from coda_mdstore.models import Bag, Bag_Info, Node, External_Identifier
-from django.core.urlresolvers import resolve
-from coda_mdstore.views import prepare_graph_date_range
+import os
+import unittest
+
 from datetime import date, datetime
```
somexpert: Think the blank line needs to go after this import, rather than before.

somexpert:
Do we really need to do the `from coda_mdstore.views import *`?

damonkelley:
@somexpert, I merely reorganized the imports and removed anything that was not used. That change is more involved and will be completed when the tests under go heavier refactoring. 

The goal for this PR is just to get tests working.

somexpert:
Well in that case, I approve this PR :thumbsup:

ldko:
:+1:

PR #16, opened 2015-07-06
[RFC] Review DB Mocking

damonkelley:
Would you all take a look at `coda/coda_mdstore/tests/test_views.py` and let me know what you think about this approach.

https://github.com/unt-libraries/coda/pull/16/files#diff-86ad8b8225379ee56381d5b558d6ceb8

Basically, what I am doing here is sandboxing a way of mocking the models that interact with the database so that I can get a better *unit* test for the views.

A new PR will be opened for merging the reorganized tests into master.

### DIFF
--- a/coda/coda_mdstore/tests/test_urls.py
+++ b/coda/coda_mdstore/tests/test_urls.py
@@ -1,0 +1,81 @@
+from django.core.urlresolvers import resolve
+
+from .. import views
+
+
+def test_index():
+    assert resolve('/').func == views.index
+
+
+def test_all_bags():
+    assert resolve('/bag/').func == views.all_bags
+
+
+def test_app_bag_no_parameters():
+    assert resolve('/APP/bag/').func == views.app_bag
+
+
+def test_app_with_parameters():
+    assert resolve('/APP/bag/ark:/67531/coda2/').func == views.app_bag
+
+
+def test_bagHTML():
+    assert resolve('/bag/ark:/67531/coda2/').func == views.bagHTML
+
+
+def test_bagURLList():
+    assert resolve('/bag/ark:/67531/coda2.urls').func == views.bagURLList
+
+
+def test_stats():
+    assert resolve('/stats/').func == views.stats
+
+
+def test_json_stats():
+    assert resolve('/stats.json').func == views.json_stats
+
+
+def test_app_node():
+    assert resolve('/APP/node/').func == views.app_node
+
+
+def test_app_node_with_identifier():
+    assert resolve('/APP/node/coda-123/').func == views.app_node
+
+
+def test_showNodeStatus():
+    assert resolve('/node/').func == views.showNodeStatus
+
+
+def test_showNodeStatus_with_identifier():
+    assert resolve('/node/coda-123/').func == views.showNodeStatus
+
+
+def test_externalIdentifierSearch_with_identifier():
+    url = resolve('/extidentifier/test_value/')
+    assert url.func == views.externalIdentifierSearch
+
+
+def test_externalIdentifierSearch():
+    url = resolve('/extidentifier/')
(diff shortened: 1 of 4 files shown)


damonkelley:
ping @somexpert, @vphill, @ldko

damonkelley:
And here is the view that is being tested 
https://github.com/unt-libraries/coda/blob/reorganize-mdstore-tests/coda/coda_mdstore/views.py#L130-L158

somexpert:
Just my two cents: I like the idea of mocking the database hits, and I think there would be cases where it would be really useful, but I don't know that I would personally want to do that each time I was writing unit tests for my views. I'm just not sure that the focus you get with this approach will _in general_ be worth the time spent writing extra code to avoid hitting the database. I realize that unit tests are meant to test as little as possible, but on the other hand, duplication in test reach is not as much of a concern as it is in the actual production code. I think if the query is complicated, or if you're going to touch many individual entries, or if you have a whole lot of tests dealing with your views, then this would be a great way to do it. But if your tests are extremely simple and minimal in number, then I'm not sure I would choose this approach. Of course, this is just the opinion of someone who has not spent a great deal of time writing tests, and so I haven't come across very confusing messages where I was unable to pinpoint the source of the problem due to unit tests not being segregated enough. If I had, perhaps my opinion would be different.

damonkelley:
>I realize that unit tests are meant to test as little as possible, but on the other hand, duplication in test reach is not as much of a concern as it is in the actual production code.

@somexpert, note that the long function is just a fixture that is available to all of the other test methods, so in the example, we are only setting up the tests once.

damonkelley:
One of the reasons I am looking at this approach is because of the how _fat_ the views are. The actual setup for some of these views if we use the database could be much longer.

somexpert:
Hmmm, yeah, I didn't really take that into consideration. Yeah, that's very true, and in that case, yeah, mocking the DB hits would make the tests more focused and faster, and any errors would of course be easier to track down.

damonkelley:
>mocking the DB hits would make the tests more focused and faster,

oh boy, they are __so__ much faster

ldko:
So far, it looks fine to me.

damonkelley:
I guess what I wondering about is how easy it would be for someone else to jump into updating/adding tests in this manner.

Is the pay off worth, the (slightly) higher cost of entry?

ldko:
In that we have to mock parts of the database rather than put entries in a test database?

damonkelley:
yes, plus the manner in which they are setup (i.e. fixtures, monkeypatching)

damonkelley:
Also, I would also like to add a few _functional_ tests at some point that would test with a real db

ldko:
I think in general it is normal for us to need to mock and patch when we are writing unit tests. I don't see monkeypatching and marking things with pytest.fixture as particularly more work than what we do for other test writing.

somexpert:
It's very likely that I'm just not a fan of mocking and patching because I haven't used them enough.

ldko:
@damonkelley  if you are ever writing any code (including these tests) where you feel like it is getting to the point of not being easily understood/worked on by others, than you should probably go with that feeling.

damonkelley:
@ldko, I think that is the answer I am trying to get by opening this PR. If you or @somexpert were to pick up the tests, would you understand what is going on? (Of course would be added to make sense of some of it)

somexpert:
Well, as for me, I can say that although the use of mocking and patching is unfamiliar, your code isn't complicated and looks like it would be easy enough to duplicate, at least within the same project.

ldko:
As far as what I see now, I think we understand it and could add tests similarly.

damonkelley:
ok cool.

Thanks for the input @somexpert @ldko. I am going to close this now. A complete PR for the `coda_mdstore` tests will follow in the next few days.

Issue #4, opened 2015-07-01
coda_mdstore Tests


damonkelley:
Add tests for `coda_mdstore`.

PR #26, opened 2015-07-31
MDStore Tests for Utilities

damonkelley:
This provides tests for the following functions:

- `coda_mdstore.views.paginage_entries`
- `coda_mdstore.views.percent`
- `coda_mdstore.views.bagFullTextSearch`
- `coda_mdstore.presentation.bagSearch`
- `coda_mdstore.presentation.makeBagAtomFeed`
- `coda_mdstore.presentation.objectsToXML`
- `coda_mdstore.presentation.getValueByName`
- `coda_mdstore.presentation.getNodeByName`
- `coda_mdstore.presentation.nodeEntry`
- `coda_mdstore.presentation.updateNode`
- `coda_mdstore.presentation.createNode`
- `coda_mdstore.presentation.xmlToBagObject`
- `coda_mdstore.presentation.createBag`
- `coda_mdstore.presentation.updateBag`

This **does not** include tests for:
- `coda_mdstore.presentation.getFileList`
    - "Untestable"*
- `coda_mdstore.presenstation.getFileHandle`
    - "Untestable"*
- `coda_mdstore.presentation.getERC`
    - will be removed
- `coda_mdstore.presentation.getERCSupport`
    - will be removed
- `coda_mdstore.presentation.load_tag_buf`
    - will be removed
- `coda_mdstore.presentation.baginfoToXHTML`
    - will be removed

Issue: #4 

---
* Not that we couldn't actually write a test for this function, but, because the function is so complex, and setting up the test would just as, or more, complex, I am going to claim that it is "untestable"

### DIFF
--- a/coda/coda_mdstore/tests/test_utils.py
+++ b/coda/coda_mdstore/tests/test_utils.py
@@ -487,6 +487,9 @@ def bag_xml():
 @pytest.mark.django_db
 @pytest.mark.usefixture('bag_xml')
 class TestCreateBag:
+    """
+    Tests for coda_mdstore.presentation.createBag.
+    """
 
     @pytest.mark.xfail(reason='Exception in function will never be raised.')
     def test_raises_exception_when_xml_cannot_be_parsed(self):
@@ -515,6 +518,9 @@ class TestCreateBag:
 @pytest.mark.django_db
 @pytest.mark.usefixture('bag_xml')
 class TestUpdateBag:
+    """
+    Tests for coda_mdstore.presentation.updateBag.
+    """
     CODA_XML = '{http://digital2.library.unt.edu/coda/bagxml/}codaXML'
 
     def test_returns_bag(self, bag_xml, rf):
--- a/coda/coda_mdstore/tests/test_utils.py
+++ b/coda/coda_mdstore/tests/test_utils.py
@@ -153,3 +153,57 @@ class TestGetNodeByName:
         node = presentation.getNodeByName(person_xml, name)
         assert node.tag == name
         assert node.text == value
+
+
+class TestNodeEntry:
+    """
+    Tests for coda_mdstore.presentation.nodeEntry.
+    """
+
+    def test_xml_has_node_attributes(self):
+        node = factories.NodeFactory.build()
+        tree = presentation.nodeEntry(node)
+        xml_str = etree.tostring(tree)
+
+        xml_obj = objectify.fromstring(xml_str)
+        assert xml_obj.content.node.name == node.node_name
+        assert xml_obj.content.node.capacity == node.node_capacity
+        assert xml_obj.content.node.size == node.node_size
+        assert xml_obj.content.node.path == node.node_path
+        assert xml_obj.content.node.url == node.node_url
+        assert xml_obj.content.node.last_checked == str(node.last_checked)
+        assert xml_obj.content.node.countchildren() == 6
+
+    def test_xml_id(self):
+        node = factories.NodeFactory.build()
+        web_root = 'example.com'
+
+        tree = presentation.nodeEntry(node, web_root)
+        xml_str = etree.tostring(tree)
+        xml_obj = objectify.fromstring(xml_str)
+
+        assert web_root in xml_obj.id.text
+
+    def test_xml_title(self):
+        node = factories.NodeFactory.build()
+        tree = presentation.nodeEntry(node)
+        xml_str = etree.tostring(tree)
+
+        xml_obj = objectify.fromstring(xml_str)
+        assert xml_obj.title == node.node_name
+
+    def test_xml_has_author_name_element(self):
+        node = factories.NodeFactory.build()
+        tree = presentation.nodeEntry(node)
+        xml_str = etree.tostring(tree)
+
+        xml_obj = objectify.fromstring(xml_str)
+        assert hasattr(xml_obj.author, 'name')
+
+    def test_xml_has_author_uri_element(self):
+        node = factories.NodeFactory.build()
+        tree = presentation.nodeEntry(node)
+        xml_str = etree.tostring(tree)
+
+        xml_obj = objectify.fromstring(xml_str)
+        assert hasattr(xml_obj.author, 'uri')
--- a/coda/coda_mdstore/tests/test_utils.py
+++ b/coda/coda_mdstore/tests/test_utils.py
@@ -69,8 +69,7 @@ class TestMakeBagAtomFeed:
         assert feed.link.get('href') == _id
         assert feed.link.get('rel') == 'self'
 
-    @mock.patch('coda_mdstore.presentation.wrapAtom')
-    def test_without_bag_objects(self, wrapAtom):
+    def test_without_bag_objects(self):
         title = 'test title'
         _id = 'test-id'
         bag_list = []
@@ -78,9 +77,9 @@ class TestMakeBagAtomFeed:
         result = presentation.makeBagAtomFeed(bag_list, _id, title)
         feed = objectify.fromstring(etree.tostring(result))
 
-        assert wrapAtom.call_count == 0
         assert feed.id == _id
         assert feed.title == title
         assert feed.updated.text is None
         assert feed.link.get('href') == _id
         assert feed.link.get('rel') == 'self'
+        assert feed.countchildren() == 4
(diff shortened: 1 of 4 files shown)


damonkelley:
ping @somexpert @ldko

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+        assert feed.updated.text is None
+        assert feed.link.get('href') == _id
+        assert feed.link.get('rel') == 'self'
+        assert feed.countchildren() == 4
+
+
+@pytest.mark.django_db
+class TestObjectsToXML:
+    """
+    Test for coda_mdstore.presentation.objectsToXML.
```
ldko: "Tests for" to be consistent :

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+
+    def test_bag_attribute_conversion(self):
+        bag = factories.FullBagFactory.create()
+        xml = etree.tostring(presentation.objectsToXML(bag))
+        bag_xml = objectify.fromstring(xml)
+
+        assert bag_xml.name == bag.name
+        assert bag_xml.fileCount == bag.files
+        assert bag_xml.payloadSize == bag.size
+        assert bag_xml.payloadSize == bag.size
```
ldko: Duplicate assert. Maybe assert was meant to be for bagitVersion.

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+
+
+@pytest.mark.django_db
+class TestObjectsToXML:
+    """
+    Test for coda_mdstore.presentation.objectsToXML.
+    """
+
+    def test_bag_attribute_conversion(self):
+        bag = factories.FullBagFactory.create()
```
ldko: Since you set up the same way for both tests in this class, should it be broken out?

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+        ('name', 'John Doe'),
+        ('age', '34')
+    ])
+    def test_returns_node(self, name, value, person_xml):
+        node = presentation.getNodeByName(person_xml, name)
+        assert node.tag == name
+        assert node.text == value
+
+
+class TestNodeEntry:
```
ldko: Lots of duplicate set up that could be moved out to a setup class.

coda/coda_mdstore/tests/test_utils.py
damonkelley: I like having them separate because they test different things, event though the setup is the same. Ideally, these would be tests for different two different functions, but ATM that is not the case.

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,572 @@
...
+        assert bag.files == str(bag_xml.fileCount)
+        assert error is None
+
+    def test_payloadSize_not_set(self, bag_xml):
+        del bag_xml.payloadSize
+        bag, bag_infos, error = presentation.xmlToBagObject(bag_xml)
+
+        assert bag is None
+        assert bag_infos is None
+        assert "Unable to set 'size' attribute" in error
```
somexpert: Why does the style of this assertion differ from those in `test_fileCount_not_set` and `test_name_not_set`?

damonkelley: So unlike the section that report an error under a certain scenario, when the `payloadSize` is retrieved, if an error is caught, that error message is appended to the `error` return variable. 

Here is the [line](https://github.com/unt-libraries/coda/blob/master/coda/coda_mdstore/presentation.py#L219)

Since we don't care about the exceptions message, I chose to only test that the first part of the message is present.

somexpert: Ah, ok, that makes sense.

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+
+        assert bag.bagit_version == str(bag_xml.bagitVersion)
+        assert error is None
+
+    def test_baggingDate_not_set(self, bag_xml):
+        bag, bag_infos, error = presentation.xmlToBagObject(bag_xml)
+
+        assert isinstance(bag.bagging_date, datetime)
+        assert error is None
+
```
ldko: Did you mean to have a test_baggingDate_is_set?

damonkelley: No I left it out intentionally.

**Why, you ask?**
I have a personal preference to avoid testing the state of datetime objects. Perhaps, I should get over it.

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+    def test_has_bag_info_objects(self, bag_xml):
+        bag, bag_infos, error = presentation.xmlToBagObject(bag_xml)
+        assert len(bag_infos) == 2
+        # Verify that all of the bag_infos are instances of models.Bag_Info
+        assert all([isinstance(m, models.Bag_Info) for m in bag_infos])
+
+    def test_has_no_bag_info_objects(self, bag_xml):
+        del bag_xml.bagInfo.item
+        del bag_xml.bagInfo.item
```
somexpert: duplicate?

coda/coda_mdstore/tests/test_utils.py
ldko: I asked because you have is set/not sets for the others including for the _very similarly_ tested `last_verified_date` where you tested for an instance of datetime object for set and not set.

damonkelley: that is a good point. i will add in `test_baggingDate_is_set`

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+    def test_raises_exception_when_xml_cannot_be_parsed(self):
+        with pytest.raises(Exception) as e:
+            presentation.createBag('')
+        assert str(e) == 'Unable to parse uploaded XML'
+
+    @pytest.mark.xfail(reason='Exception in function will never be raised.')
+    def test_raises_exception_when_content_element_not_present(self):
+        with pytest.raises(Exception) as e:
+            presentation.createBag('<root/>')
+        assert str(e) == 'No content to parse uploaded XML'
```
ldko: Where did this string come from? Should it be "No content element located"?

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -1,14 +1,573 @@
...
+        xml_str = etree.tostring(bag_xml)
+        created_bag, created_bag_infos = presentation.createBag(xml_str)
+        assert isinstance(created_bag, models.Bag)
+
+    def test_returns_bag_info_objects(self, bag_xml, rf):
+        xml_str = etree.tostring(bag_xml)
+        created_bag, created_bag_infos = presentation.createBag(xml_str)
+        for bag_info in created_bag_infos:
+            assert isinstance(bag_info, models.Bag_Info)
+
```
ldko: Should there be an additional test in this class for:  If a mocked xmlToBagObject returns an error_code?
Also. what is the feasibility of testing this:
    codaXML = contentElement.xpath("*[local-name() = 'codaXML']")[0]
    codaName = codaXML.xpath("*[local-name() = 'name']")[0].text.strip()
    oldBagInfoObjectList = Bag_Info.objects.filter(bag_name=codaName)
    for oldBagInfoObject in oldBagInfoObjectList:
        oldBagInfoObject.delete()

coda/coda_mdstore/tests/test_utils.py
damonkelley: It is a test for this exception

[Code](https://github.com/unt-libraries/coda/blob/master/coda/coda_mdstore/presentation.py#L172)

ldko: I thought it was at test for this [exception] (https://github.com/unt-libraries/coda/blob/master/coda/coda_mdstore/presentation.py#L175) and the code you linked is tested by the previous method.

coda/coda_mdstore/tests/test_utils.py
damonkelley: 1. yes
2. Feasible? yes. Pretty? i doubt it.

ldko: TestUpdateBag could also have a test for deleting old bag infos.

damonkelley: complete in 2b01b2b

somexpert:
:thumbsup:

coda/coda_mdstore/tests/test_utils.py
ldko: I am still wondering if this is right.

ldko:
:+1: cha ching

PR #28, opened 2015-08-04
Tests for the mdstore feed pagination.

damonkelley:
Issue: #4

### DIFF
--- a/coda/coda_mdstore/tests/test_feeds.py
+++ b/coda/coda_mdstore/tests/test_feeds.py
@@ -1,0 +1,54 @@
+from lxml import objectify
+import pytest
+
+from django.core.paginator import EmptyPage
+
+from .. import factories
+
+
+@pytest.mark.django_db
+class TestAtomSiteNewsFeed:
+
+    def test_num_entries_per_page(self, client):
+        factories.BagFactory.create_batch(30)
+        response = client.get('/feed/')
+
+        feed = objectify.fromstring(response.content)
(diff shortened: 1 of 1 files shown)


damonkelley:
ping @ldko @somexpert

coda/coda_mdstore/tests/test_feeds.py
```diff
@@ -0,0 +1,54 @@
...
+        assert feed.link[-1].get('rel') == 'last'
+
+    def test_has_previous_related_link(self, client):
+        factories.BagFactory.create_batch(30)
+        response = client.get('/feed/', {'p': 2})
+
+        feed = objectify.fromstring(response.content)
+
+        assert feed.link[-1].get('rel') == 'previous'
+        assert '?p=1' in feed.link[-1].get('href')
```
somexpert: This is extremely minor, but can we swap these two lines to match up with the previous tests?

coda/coda_mdstore/tests/test_feeds.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_invalid_page_raises_exception(self, client):
```
ldko: Should we change this to a test than we can expect to fail--or do we really want to raise an exception when the page parameter is for a page that doesn't exist?

somexpert:
:thumbsup:

ldko:
:+1:

Issue #6, opened 2015-07-01
coda_replication Tests


damonkelley:
Add tests for `coda_replication`.

PR #32, opened 2015-08-10
Replication view tests

damonkelley:
Partially completes issue: #6

### DIFF
--- a/coda/coda_replication/factories.py
+++ b/coda/coda_replication/factories.py
@@ -1,0 +1,24 @@
+"""
+Coda Replication Model factories for test fixtures.
+"""
+from datetime import datetime
+
+import factory
+from factory import fuzzy
+
+from . import models
+
+
+class QueueEntryFactory(factory.django.DjangoModelFactory):
+
+    class Meta:
+        model = models.QueueEntry
+
+    ark = factory.Sequence(lambda n: 'ark:/00001/id{0}'.format(n))
+    bytes = fuzzy.FuzzyInteger(100000000)
+    files = fuzzy.FuzzyInteger(50, 500)
+    url_list = fuzzy.FuzzyText(length=500)
+    status = fuzzy.FuzzyChoice(str(i) for i in range(1, 10))
+    harvest_start = fuzzy.FuzzyNaiveDateTime(datetime(2015, 01, 01))
+    harvest_end = fuzzy.FuzzyNaiveDateTime(datetime(2015, 06, 01))
+    queue_position = fuzzy.FuzzyInteger(1, 100)
--- a/coda/coda_replication/factories.py
+++ b/coda/coda_replication/factories.py
@@ -10,10 +10,6 @@ from . import models
 
 
 class QueueEntryFactory(factory.django.DjangoModelFactory):
-
-    class Meta:
-        model = models.QueueEntry
-
     ark = factory.Sequence(lambda n: 'ark:/00001/id{0}'.format(n))
     bytes = fuzzy.FuzzyInteger(100000000)
     files = fuzzy.FuzzyInteger(50, 500)
@@ -22,3 +18,6 @@ class QueueEntryFactory(factory.django.DjangoModelFactory):
     harvest_start = fuzzy.FuzzyNaiveDateTime(datetime(2015, 01, 01))
     harvest_end = fuzzy.FuzzyNaiveDateTime(datetime(2015, 06, 01))
     queue_position = fuzzy.FuzzyInteger(1, 100)
+
+    class Meta:
+        model = models.QueueEntry
--- a/coda/coda_replication/tests/test_views.py
+++ b/coda/coda_replication/tests/test_views.py
@@ -471,7 +471,7 @@ class TestQueue:
 
         assert response.status_code == 404
 
-    @pytest.mark.xfail(reason="The updateQueueEntry subroutine causes an exception to be trown.")
+    @pytest.mark.xfail(reason="The updateQueueEntry subroutine causes an exception to be thrown.")
     def test_put_with_xml_for_entry_that_does_not_exist(self, queue_xml, rf):
         request = rf.put('/', queue_xml, 'application/xml', HTTP_HOST='example.com')
         response = views.queue(request, 'ark:/67531/coda4fnk')
--- a/coda/coda_replication/tests/test_views.py
+++ b/coda/coda_replication/tests/test_views.py
@@ -309,6 +309,9 @@ class TestQueueRecent:
 
 
 class TestQueue:
+    """
+    Tests for coda_replication.views.queue.
+    """
 
     @pytest.fixture
     def queue_xml(self):
@@ -348,6 +351,7 @@ class TestQueue:
 
         xml = objectify.fromstring(response.content)
         assert len(xml.entry) == 10
+        assert response.status_code == 200
         assert response['Content-Type'] == 'application/atom+xml'
 
     def test_get_without_identifier_with_not_entries(self, rf):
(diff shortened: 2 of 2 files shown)


damonkelley:
Ping @somexpert @ldko

coda/coda_replication/factories.py
```diff
@@ -0,0 +1,23 @@
+"""
+Coda Replication Model factories for test fixtures.
+"""
+from datetime import datetime
+
+import factory
+from factory import fuzzy
```
somexpert: Can we join these two and just do something like `from factory import fuzzy, Sequence` and then change line 13 to be `ark = Sequence(lambda n: 'ark:/00001/id{0}'.format(n))`?

damonkelley: No because we would have to name every import that we need from the `factory` module. The only reason `fuzzy` is imported is because the package requires an explicit import. See [here](https://factoryboy.readthedocs.org/en/latest/fuzzy.html#module-factory.fuzzy)

coda/coda_replication/tests/test_views.py
```diff
@@ -0,0 +1,406 @@
...
+        factories.QueueEntryFactory.create_batch(30)
+        request = rf.get('/', HTTP_HOST='example.com')
+        response = views.queue(request)
+
+        xml = objectify.fromstring(response.content)
+        assert len(xml.entry) == 10
+        assert response.status_code == 200
+        assert response['Content-Type'] == 'application/atom+xml'
+
+    def test_get_without_identifier_with_not_entries(self, rf):
```
somexpert: Should the "with_not" perhaps be changed to "without" or "with_no"?

somexpert:
Doesn't seem like queue_html() gets tested. Should it?

coda/coda_replication/tests/test_views.py
```diff
@@ -0,0 +1,406 @@
...
+            reverse('coda_replication.views.queue_search'),
+            {'status': status})
+
+        results = response.context[-1]['entries']
+        assert sum(q.status == status for q in entries) == len(results.object_list)
+
+    def test_sorting_by_size(self, client):
+        """
+        Tests that the QueueEntry search results are ordered by the bytes
+        attribute in ascending order with the sort query is `size`.
```
somexpert: 'when' instead of 'with' perhaps?

coda/coda_replication/tests/test_views.py
```diff
@@ -0,0 +1,406 @@
...
+            reverse('coda_replication.views.queue_search'),
+            {'sort': 'size'})
+
+        results = response.context[-1]['entries']
+        entries.sort(key=lambda x: x.bytes)
+
+        for entry, result in zip(entries, results.object_list):
+            assert entry.bytes == result.bytes
+
+    def test_filtering_by_harvest_end(self, client):
```
somexpert: Since this and the next test are testing filtering out results based on a date, wouldn't it be good to have a model instance which should not be grabbed by the query, just to be sure that you're getting back only what you want?

damonkelley: `end_date` is not actually not an instance of `QueueEntry`, rather it is a datetime object. Instead of actually testing the result of the filter, this really verifies if the 'end_date` branch of the view is executed. The only time `context['end_date`]` is set is if that branch is executed. 

I don't love it, but it is better than checking if all of the models in the db >= a datetime object are present in the response context. :-1: 

To tell you the truth, I don't really like the idea of testing any of the sorting/filtering on the _search_ views.

somexpert:
In the TestQueue class, should we also have a `test_get_with_identifier_with_no_entries()` to test that `except` branch?

somexpert:
:thumbsup:

ldko:
Things we might want to add tests for:
1. test_template_used test in TestQueueSearch and TestQueueHtml for consistency?
2. We might want a test in TestQueue for a PUT request on an entry that doesn't exist. Currently we get a 500 error, but we should probably be creating it and returning a 201 (https://github.com/unt-libraries/coda/issues/33).
3. Our harvester generally makes requests to /APP/queue with specific status and count parameter values, so it would be good to have tests for those conditions where a non-default status and count are used.
4. TestQueue throws 500 error if page parameter value is for a non-existent page, though problem is in bagatom (https://github.com/unt-libraries/codalib/issues/6).
5. TestQueue throws 500 error if you have a sort parameter without a status parameter because if you don't, it tries to sort something that doesn't exist (https://github.com/unt-libraries/coda/issues/34). We should have a failing test here.

ldko:
Regarding number 2 in my previous comment, I had said we should create a resource that doesn't exist if being sent with a PUT because of the [AtomPub spec] (http://bitworking.org/projects/atom/rfc5023.html#edit-via-PUT) saying:
```
 "To edit a Member Resource, a client sends a PUT request to its Member URI,
 as specified in [RFC2616]."
```
which talks about creating a resource that is being PUT if it doesn't exist.

But [later in the AtomPub spec] (http://bitworking.org/projects/atom/rfc5023.html#codeisdata) it says:
```
 "PUT is used to edit a known Resource. It is not used for Resource creation."
```
So perhaps we don't need a test that would expect a 201, but possibly a 404.

damonkelley:
>Things we might want to add tests for:
1. test_template_used test in TestQueueSearch and TestQueueHtml for consistency?
2. We might want a test in TestQueue for a PUT request on an entry that doesn't exist. Currently we get a 500 error, but we should probably be creating it and returning a 201 (https://github.com/unt-libraries/coda/issues/33).
3. Our harvester generally makes requests to /APP/queue with specific status and count parameter values, so it would be good to have tests for those conditions where a non-default status and count are used.
4. TestQueue throws 500 error if page parameter value is for a non-existent page, though problem is in bagatom (https://github.com/unt-libraries/codalib/issues/6).
5. TestQueue throws 500 error if you have a sort parameter without a status parameter because if you don't, it tries to sort something that doesn't exist (https://github.com/unt-libraries/coda/issues/34). We should have a failing test here.


@ldko, these are addressed in e8d73be...a0fd007

coda/coda_replication/tests/test_views.py
```diff
@@ -0,0 +1,496 @@
...
+
+    @pytest.mark.xfail(reason="The PUT method with out an identifier should not be allowed.")
+    def test_put_without_identifier(self, queue_xml, rf):
+        factories.QueueEntryFactory.create(ark='ark:/67531/coda4fnk', bytes='4')
+        request = rf.put('/', queue_xml, 'application/xml', HTTP_HOST='example.com')
+        response = views.queue(request)
+
+        assert response.status_code == 404
+
+    @pytest.mark.xfail(reason="The updateQueueEntry subroutine causes an exception to be trown.")
```
ldko: "trown" typo

coda/coda_replication/tests/test_views.py
```diff
@@ -0,0 +1,496 @@
...
+        assert response.status_code == 200
+        assert response['Content-Type'] == 'application/atom+xml'
+
+    @pytest.mark.xfail(reason="The PUT method with out an identifier should not be allowed.")
+    def test_put_without_identifier(self, queue_xml, rf):
+        factories.QueueEntryFactory.create(ark='ark:/67531/coda4fnk', bytes='4')
+        request = rf.put('/', queue_xml, 'application/xml', HTTP_HOST='example.com')
+        response = views.queue(request)
+
+        assert response.status_code == 404
```
ldko: Instead of a 404, would we be expecting a 405?

damonkelley: hard to say. Some of this stuff will likely change quite a bit. Ill add it in for now at least

ldko:
:+1:

Issue #7, opened 2015-07-01
coda_validate Tests


damonkelley:
Add tests for `coda_validate`.

PR #38, opened 2015-08-18
Replication Utilities Tests

damonkelley:
Will partially complete issue #7 

Notes the function `queue_object_to_xml` does not have tests because it is never used.


damonkelley:
ping @ldko @somexpert

ldko:
:+1: (but that position vs. queue_position is sad)

somexpert:
:thumbsup:

damonkelley:
@ldko at least it is a super easy fix. (#39)

PR #40, opened 2015-08-20
Validate Utilities Tests

damonkelley:
Partially completes issue #7


damonkelley:
ping @ldko @somexpert

coda/coda_validate/tests/test_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    assert validate_xml.last_verified == str(validate.last_verified)
+    assert validate_xml.last_verified_status == str(validate.last_verified_status)
+    assert validate_xml.priority_change_date == str(validate.priority_change_date)
+    assert validate_xml.priority.text == validate.priority
+    assert validate_xml.server == validate.server
+
+
+@pytest.mark.xfail(reason='Exception will never be raised directly from the function.')
+def test_xmlToValidateObject_raises_exception():
+    assert 0
```
somexpert: Shouldn't there be test here?

damonkelley: this is more to document that something needs to change about this function rather than to create a test that will pass in the future.

(we discussed this in person. just putting it here for @ldko)

somexpert:
:thumbsup:

coda/coda_validate/tests/test_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+    factories.ValidateFactory.create(
+        identifier=validate_xml.identifier,
+        priority=validate_xml.priority
+    )
+
+    updated_validate = views.xmlToUpdateValidateObject(validate_feed)
+
+    assert validate_xml.priority > 0
+    assert updated_validate.priority is 0
```
damonkelley: Change this to

```python
assert updated_validate.priority == 0
```

Be explicit. We are checking the value, not the identity.

ldko:
:+1:

PR #44, opened 2015-08-27
Tests for AtomNextNewsFeed.

damonkelley:
Partially resolves issue #7

### DIFF
--- a/coda/coda_validate/tests/test_feeds.py
+++ b/coda/coda_validate/tests/test_feeds.py
@@ -1,0 +1,64 @@
+from datetime import datetime
+
+import pytest
+
+from django.conf import settings
+
+from .. import views, factories
+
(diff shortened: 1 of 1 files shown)


damonkelley:
ping @ldko @somexpert

damonkelley:
Once this and #45 are merged the validate app will have sufficient test coverage.

somexpert:
While I'm not a big fan of testing the reason string but not the object being returned, I can totally see why you would do that. Plus, the feed code is flagelous (as we discussed before), so :thumbsup:

damonkelley:
>While I'm not a big fan of testing the reason string but not the object being returned, I can totally see why you would do that.

I am right there with you.

ldko:
:+1:

PR #45, opened 2015-08-27
Test for Validate Models

damonkelley:
Partially completes issue #7


damonkelley:
ping @ldko @somexpert

somexpert:
Works for me. Not much to test in that model. :thumbsup:

ldko:
:+1:

PR #47, opened 2015-09-10
Resource-Sync Tests

damonkelley:
Will partially resolve issue #4 

Note that I have intentionally not included comprehensive tests for `index` and `sitemap` because they are copied almost line for line from Django source, which is already testing these functions. I have opened issue #48 for removing this code duplication.

### DIFF
--- a/coda/coda_mdstore/tests/test_resourcesync.py
+++ b/coda/coda_mdstore/tests/test_resourcesync.py
@@ -1,0 +1,66 @@
+import pytest
+
+from django.contrib.sitemaps import Sitemap
+
+from .. import factories
+from .. import resourcesync
+
+pytestmark = [
+    pytest.mark.urls('coda_mdstore.urls'),
+    pytest.mark.django_db()
+]
+
+
+class TestSitemap(Sitemap):
+    lastmod = None
+    protocol = 'http'
+
+    def items(self):
+        return factories.FullBagFactory.build_batch(5001)
+
+    def location(self, obj):
+        return "/bag/%s" % obj.name
+
+
+def test_index(rf):
+    sitemaps = {'001': TestSitemap}
+    request = rf.get('/')
+    response = resourcesync.index(request, sitemaps, 'mdstore/resourceindex.xml')
+
+    content = response.render()
+
+    locations = [
+        'resourcelist-001.xml',
+        'resourcelist-002.xml'
+    ]
+
+    for loc in locations:
+        assert loc in str(content)
+
+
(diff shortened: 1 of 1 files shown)


damonkelley:
ping @ldko @somexpert

ldko:
Consider adding some comments. Otherwise, :+1:

coda/coda_mdstore/tests/test_resourcesync.py
```diff
@@ -0,0 +1,167 @@
...
+    """Fake Sitemap object similar to resourcesync.BaseSitemap.
+
+    This will build but not persist model instances for testing
+    purposes.
+    """
+    lastmod = None
+    protocol = 'http'
+
+    def items(self):
+        # A miniumum of 50001 Bag objects are required to get multiple
```
somexpert: One too many zeros

coda/coda_mdstore/tests/test_resourcesync.py
```diff
@@ -0,0 +1,167 @@
...
+
+    request = rf.get('/')
+    response = resourcesync.sitemap(request, resourcesync.sitemaps, 1, 'mdstore/sitemap.xml')
+    response.render()
+
+    # Verify that that the following urls are present for each bag in the
+    # bags batch.
+    for bag in bags:
+        assert 'http://example.com/bag/{0}'.format(bag.name) in response.content
+        assert 'http://example.com/bag/{0}.urls'.format(bag.name) in response.content
```
somexpert: These two assertions seem problematic. The first assertion is looking for essentially a prefix of the second. So both assertions will always be true if the second string is present.

coda/coda_mdstore/tests/test_resourcesync.py
ldko: I do particularly like comments like this where it indicates why you used `5001` more than comments like "Check that all the dictionaries in `urlset` have the following keys" which more just say exactly what the code says (and that you probably would have thought too redundant to add if I hadn't mentioned adding comments). :)

coda/coda_mdstore/tests/test_resourcesync.py
damonkelley: The point was to be explicit. By asserting that both urls are present the reader gets a feel for what the function under test is doing and the nature of its return value.

However, I really don't like that the content is being tested. We don't need to involve the template just to test the view. I am going to rework this to the only test the context.

somexpert:
:thumbsup:

PR #62, opened 2015-12-16
Staticfiles

### LINKED ISSUE #59: Update to use STATIC_URL
Previously, the project used `MEDIA_URL` as the url prefix for retrieving the staticfiles. We need to update this to `STATIC_URL` and replace all instances of `MEDIA_URL` in the application code with calls to `static`. 

Part of completing this ticket should include fixing the Admin assets which broke after #56.

damonkelley:
This removes the `MEDIA_URL` setting and replaces it with the `django.contrib.staticfiles` app, which uses the `STATIC_URL` setting.

Resolves #59


damonkelley:
ping @ldko @somexpert

somexpert:
:thumbsup:

ldko:
:+1:

PR #63, opened 2015-12-17
Upgrade URLs

### LINKED ISSUE #14: Update the mdstore url patterns
The urls in `coda_mdstore` are super old school. They will need to be upgraded to use the `url()` function. Make sure we also give them `names` and set the namespace to `mdstore`

Once that is done, update the url test names to use the name of the url instead of the name of the view it is resolving to.

damonkelley:
Resolves #14, but also updates the urls for the other apps.

Essentially this converts the old style urls

```python
urlpatterns = patterns(''
    (r'^pattern', 'python.path.to.view'),
    # ....
)
```

To the new style

```python
urlpatterns = [
    url(r'^pattern', python.path.to.view, name='some-url'),
    # ....
]
```

Note that I chose to leave the `resourcesync` URLs as is. I attempted to change them, there was an issue with the sitemaps view.

### DIFF
--- a/coda/coda_mdstore/urls.py
+++ b/coda/coda_mdstore/urls.py
@@ -1,35 +1,33 @@
-from django.conf.urls import *
-from .views import AtomSiteNewsFeed
+from django.conf.urls import url, patterns
+from . import views
 from .resourcesync import sitemaps
 
-urlpatterns = patterns(
-    'coda_mdstore.views',
-    # bag urls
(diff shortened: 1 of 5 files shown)


damonkelley:
ping @ldko @somexpert

somexpert:
:thumbsup:

ldko:
:+1:

PR #66, opened 2016-01-06
Fix response codes from utility functions

### LINKED ISSUE #23: Status Code For Invalid Node Path in updateNode
In the function `coda_mdstore.presentation.updateNode`, if the request path does not include the node name, an HttpResponse is returned without updating the node. The problem is that the status code returned with the response is 200. Since the request was not `OK` we should set a more correct status code. 

[Snippet](https://github.com/unt-libraries/coda/blob/af9e8ce800cc94aef91bc3c30844bfd1b8d7d786/coda/coda_mdstore/presentation.py#L464-L466)

damonkelley:
Fixes #23 

This modifies `coda_mdstore.presentation.updateBag` and `coda_mdstore.presentation.updateNode` to only return a Bag instance or Node instance respectively. As utility functions, they should not be directly influencing the type of the HTTP response the view sends back. Furthermore, we only want a function to return a single type of object (instead of an `HttpResponse` OR a `Bag`/`Node`).

Now `updateBag` and `updateNode` will raise exceptions in the two situations where it would previously return an `HttpResponse`.

1. If the model is not found, a `DoesNotExist`  exception will be allowed to bubble up.
2. If the name of the instance is not in the request path, a `ModelBadName` exception will be raised.

The calling controllers now know how to deal with these exceptions and will return a `404` or `400` response if encountered.

### DIFF
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -1,10 +1,7 @@
 from django.contrib.sites.models import Site
-from django.conf import settings
 from django.http import HttpResponse
 from coda_mdstore.models import Bag, Bag_Info, Node, External_Identifier
 from codalib import anvl, APP_AUTHOR
-from codalib.bagatom import wrapAtom, ATOM, ATOM_NSMAP, BAG, BAG_NSMAP, \
-    queueEntryToXML, getValueByName, getNodeByName
 from pypairtree import pairtree
 from BeautifulSoup import BeautifulSoup as BSoup
 from lxml import etree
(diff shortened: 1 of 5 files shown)


damonkelley:
ping @ldko @somexpert

somexpert:
:thumbsup:

coda/coda_mdstore/presentation.py
```diff
@@ -136,17 +137,12 @@ def updateBag(request):
...
-        resp = HttpResponse(
-            'The node name supplied in the URL does not match the XML'
-        )
-        return resp
+
+    bag = Bag.objects.get(name=bag_name)
+
+    if bag_name not in request.path:
+        raise exceptions.BadBagName('The bag name supplied in the URL does not match the XML')
+
```
ldko: Should the `bag = Bag.objects.get(name=bag_name)` line and the following if block be switched to prevent hitting the db when the URL and XML bag names don't  match? If so, same comment for bad node.

damonkelley: I think so. It is silly to do the query and then throw away the result in that scenario.

ldko:
:+1:

damonkelley:
@ldko, I made that change

ldko:
:+1:

PR #71, opened 2016-01-11
Enhancements for Docker Compose Config

damonkelley:
While I was trying to fix the Travis builds for another feature branch, I discovered a way to make running tests and the `manage.py` script much easier.

Instead of overriding the command for the `coda` container via (This is how we have been doing it)

```sh
$ docker-compose run coda ./manage.py check
```

There is now a `manage` service defined in the `docker-compose.yml` that places the `manage.py` script as the entrypoint. This means that the `manage.py` subcommands can be given directly to the service

```sh
$ docker-compose run manage check
$ docker-compose run migration
...
```

The `test` service similarly sets the `py.test` command as the entrypoint
```sh
$ docker-compose test
$ docker-compose test --create-db
```

When the build changes for the  `coda` container (now called `app`), you have to rebuild the `manage`, `app`, and `test`, because they extend from the `base` service which is built from the `Dockerfile` in the root of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,30 +29,40 @@ $ docker-compose up -d db
 
 Start the app and run the migrations.
 ```sh
-# start the app
-$ docker-compose up -d
-
 # run the migrations
-$ docker-compose run --rm coda ./manage.py migrate
+$ docker-compose run --rm manage migrate
+
+# start the app
+$ docker-compose up -d app
 
 # Optional: add a superuser in order to log in to the admin interface
-$ docker-compose run --rm web ./manage.py createsuperuser
+$ docker-compose run --rm manage createsuperuser
 ```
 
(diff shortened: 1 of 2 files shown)


damonkelley:
ping @ldko @somexpert

README.md
```diff
@@ -29,30 +29,40 @@ $ docker-compose up -d db
...
+$ docker-compose rm app test manage
 
-# rebuild the coda container
-$ docker-compose build coda
+# rebuild the app, manage, and test containers
+$ docker-compose build 
 
 # start the app
-$ docker-compose up -d
+$ docker-compose up app -d
```
ldko: should this be `docker-compose up -d app`?

ldko:
:+1:

somexpert:
:thumbsup:

PR #72, opened 2016-01-11
Initial migrations

### LINKED ISSUE #52: Add migrations
We need to get migrations in here to reflect the state of the database schema as it is now.

damonkelley:
Fixes #52 

In addition to adding initial migrations and the FULLTEXT indexes required for some search capabilities, there is also a change to set the default storage engine for all database tables to MyISAM, as this is what we are using in production.

### DIFF
--- a/coda/coda_validate/views.py
+++ b/coda/coda_validate/views.py
@@ -28,7 +28,6 @@ from codalib.bagatom import wrapAtom, makeObjectFeed
 
 
 XML_HEADER = "<?xml version=\"1.0\"?>\n%s"
-DOMAIN = Site.objects.get_current().domain
 
 
 class CorrectMimeTypeFeed(Atom1Feed):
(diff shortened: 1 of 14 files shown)


damonkelley:
ping @ldko @somexpert

damonkelley:
Disregard the ping @somexpert @ldko.

I just found more indexes that need to be included.

damonkelley:
ping @ldko @somexpert 

There was one missing index which whas `QueueEntry.bytes`. All of the indexes from models represented in this code base are now present in a migration.

coda/coda_validate/views.py
```diff
@@ -28,7 +28,6 @@
 
 
 XML_HEADER = "<?xml version=\"1.0\"?>\n%s"
-DOMAIN = Site.objects.get_current().domain
```
damonkelley: This needed to be moved into the (only) class that used it which is `CorrectMimeTypeFeed. It was causing an exception to be raised when the trying to run the migrations.

somexpert:
:thumbsup:

ldko:
:+1:

Issue #22, opened 2015-07-27
Remove unused functions


damonkelley:
This is a running list of functions that can be removed. 

- [x] `coda_mdstore.presentation.load_tag_buf`
    - Never called
- [x] `coda_mdstore.presentation.baginfoToXHTML`
   - Never called
- [x] `coda_mdstore.presentation.getERC`
    - ERC is going to be removed (#20)
- [x] `coda_mdstore.presentation.getERCSupport`
    - ERC support is going to be removed (#20)

Issue #20, opened 2015-07-24
coda_mdstore ERC support needs refactoring or removal


ldko:
In coda_mdstore/views.py, [app_bag] (https://github.com/unt-libraries/coda/blob/master/coda/coda_mdstore/views.py#L849) function has logic for `GET` requests that have an `identifier` to check if the URL has "?" or "??" appended (in support of Electronic Resource Citations (ERC) https://tools.ietf.org/html/draft-kunze-erc-01 spec).

Logic to set value for url_test in the function will exception in mod_wsgi deployments with: `AttributeError: 'WSGIRequest' object has no attribute '_req'`. Our code should not be accessing `_` prefixed attributes anyway. 

Mark's proposed solution (specifically in regard to aubrey not coda) after [this discussion] (https://groups.google.com/forum/#!topic/arks-forum/LIMI699DrdY) was:
Have a rewrite rule that would match a trailing single and double question and have it append something to the URL to let it know on the backend what to do. We would just add a erc_ark/ and erc_commitment/ view to each object--all just on the backend.

But first, do we want to maintain support for ERC in Coda?

damonkelley:
If we keep it, it might be good to put the ERC stuff in it's own app, which we can also use in Aubrey in the future.

ldko:
From Mark:
"We don't need to keep the erc code in coda."

So, we can remove ERC stuff from presentation and views.

PR #73, opened 2016-01-15
Remove ERC

damonkelley:
This takes care of #20 and #22, which removes ERC support and gets rid of some functions that are not used.


damonkelley:
ping @ldko @somexpert

coda/coda_mdstore/views.py
```diff
@@ -826,47 +828,20 @@ def app_bag(request, identifier=None):
...
-            author=APP_AUTHOR.get('name', None),
-            author_uri=APP_AUTHOR.get('uri', None)
-        )
-        entryText = XML_HEADER % etree.tostring(returnEntry, pretty_print=True)
-        resp = HttpResponse(entryText, content_type="application/atom+xml")
-        return resp
+            return HttpResponseNotFound("There is no bag with id '{0}'.".format(identifier))
+
+        object_xml = objectsToXML(bag)
+        url = request.build_absolute_uri(reverse('app-bag-detail', args=[identifier]))
```
ldko: This is causing me this error:
NoReverseMatch at /APP/bag/ark:/67531/codavefv/
Reverse for 'app-bag-detail' with arguments '(u'ark:/67531/codavefv',)' and keyword arguments '{}' not found. 0 pattern(s) tried: []

damonkelley: Interesting. 

I am not getting any errors, and it looks like the Travis build isn't either.

damonkelley: Did you rebuild all of the containers and create a new database when you ran the tests?

somexpert:
:thumbsup: (no errors on my end)

coda/coda_mdstore/views.py
damonkelley: OK, I see what is going on here.

`reverse` is not able to find the `app-bag-detail` because it is missing the namespace defined in `config/urls.py`. The reason that the tests did not find this is becuase we are overriding the main `urls.py` configuration to prevent django from loading *all* of the URLs. 

Originally, this setup was useful because we were testing the apps in isolation as we added tests and upgraded to Django 1.8. I am not sure it is still useful.

I am going to have to take another look at this configuration.

damonkelley:
@ldko, with 289220c that error you found is now covered by the tests. All of the tests now use the urls configuration at `config/urls.py` now.  I have also removed the URL namespaces.

somexpert:
:thumbsup:

ldko:
:+1:

PR #75, opened 2016-01-20
Use the correct attribute name to get the last_verified_date.

### LINKED ISSUE #21: objectsToXML Always Raises Exception
The function`coda_mdstore.presentation.objectsToXML` attempts to get the value from the `last_verified` attribute on the bag object. No such attribute exists, so an exception is always raised, and always caught. The result of this is that a `<lastVerified/>` element is added, but the value is never set (See [here](https://github.com/unt-libraries/coda/blob/af9e8ce800cc94aef91bc3c30844bfd1b8d7d786/coda/coda_mdstore/presentation.py#L288-L294))

A side effect of this bug is that the `<lastStatus/>`, which is created within the try/except, is never added and a value is never set.

damonkelley:
Fixes #21

### DIFF
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -283,7 +283,7 @@ def objectsToXML(bagObject):
     bagitVersion.text = bagObject.bagit_version
     try:
         lastVerified = etree.SubElement(codaXML, BAG + "lastVerified")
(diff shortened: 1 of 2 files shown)


damonkelley:
ping @ldko @somexpert

ldko:
:+1:

somexpert:
:thumbsup:

PR #77, opened 2016-01-22
Extract the queue list branch into a new view. Fixes #34

### LINKED ISSUE #34: coda_replication.views.queue Error if sort parameter sent without status parameter
UnboundLocalError at [queueEntries = queueEntries.order_by('bytes')] (https://github.com/unt-libraries/coda/blob/master/coda/coda_replication/views.py#L403) if a `status` parameter was not sent with `sort` parameter because in that case `queueEntries` won't exist yet.

damonkelley:
* Fixes the 500 error when status parameter is sent without a sort parementer
  as well.
* Extract the branch into a separate function that we can later map to its own
  URL pattern if we choose.
* Clean up the view and remove some unneeded code.

ping -> @ldko @somexpert

### DIFF
--- a/coda/coda_replication/views.py
+++ b/coda/coda_replication/views.py
@@ -386,52 +386,39 @@ def queue(request, identifier=None):
         )
         resp = HttpResponse(atomText, content_type="application/atom+xml")
     elif request.method == 'GET' and not identifier:
-        # return a feed of entries
-        # the value 'start' indicates the first entry, defaults to 1
-        if "page" in request.REQUEST:
-            page = int(request.REQUEST["page"])
-        else:
-            page = 1
-        # see if we have a sort value.  If not, sort by queue_position
-        if "status" in request.REQUEST:
-            # to further filter our queueEntries based on a given status
-            statusCode = request.REQUEST["status"]
-            if statusCode:
-                queueEntries = QueueEntry.objects.filter(status=statusCode)
-        if request.GET.get('sort'):
-            if request.GET['sort'] == 'size':
-                queueEntries = queueEntries.order_by('bytes')
-        else:
-            queueEntries = QueueEntry.objects.order_by("queue_position")
-        # see if we have a sort value.  If not, sort by queue_position
-        if "count" in request.REQUEST:
-            # to further filter our queueEntries based on a given status
-            count = int(request.REQUEST["count"])
-        else:
-            count = 10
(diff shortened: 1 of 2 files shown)


coda/coda_replication/views.py
```diff
@@ -386,52 +386,38 @@ def queue(request, identifier=None):
...
-            atomFeed, pretty_print=True
-        )
-        resp = HttpResponse(feedText, content_type="application/atom+xml")
-        resp.status_code = 200
+        return queue_list(request)
     return resp
+
+
+def queue_list(request):
+    status = request.GET.get('status')
```
damonkelley: A doc string wouldn't hurt here.

coda/coda_replication/views.py
```diff
@@ -386,52 +386,39 @@ def queue(request, identifier=None):
...
-        resp.status_code = 200
-    return resp
+            author=APP_AUTHOR)
+    except EmptyPage:
+        return HttpResponseBadRequest("Page does not exist.")
+
+    feedText = etree.tostring(atomFeed, pretty_print=True)
+    feedText = '<?xml version="1.0"?>\n{0}'.format(feedText)
+
+    return HttpResponse(feedText, content_type="application/atom+xml")
```
somexpert: NBD, we've already got mixed use of single quotes and double quotes elsewhere in this file, but since we're already modifying these lines, can we use single quotes in these two returns?

damonkelley: be50f45

somexpert:
:thumbsup:

coda/coda_replication/views.py
```diff
@@ -386,52 +386,39 @@ def queue(request, identifier=None):
...
+    paginator = Paginator(queues, count)
+
+    try:
         atomFeed = makeObjectFeed(
             paginator=paginator,
             objectToXMLFunction=queueEntryToXML,
             feedId=request.path[1:],
-            title="Queue Entry Feed",
+            title='Queue Entry Feed',
             webRoot='http://%s' % request.META['HTTP_HOST'],
```
ldko: Do you think we should change this string formatting to use `.format()` to be consistent with the `feedText` change made later in this function?

damonkelley: yeah. I think I just reformatted what was there without paying much attention.

ldko:
:+1:

PR #78, opened 2016-01-25
Refactor xmlToQueueEntry

### LINKED ISSUE #39: xmlToQueueEntry does not set the queue position
```python
    if position:
        entry.position = int(getValueByName(queueEntryNode, "position"))
```

Here the `position` attribute on `QueueEntry` object is being set. Unfortunately, `position` is not mapped to a table column, and the value is not being persisted. To correct this we need to change the attribute to `queue_position`.

Link to actual code: [`xmlToQueueEntry()`](https://github.com/unt-libraries/coda/blob/d0199426cd8a3a4d4b318bc8e5411c3559025a93/coda/coda_replication/presentation.py#L60)

damonkelley:
* Clean up if statements.
* Fix xfailing tests. Fixes #39.

### DIFF
--- a/coda/coda_replication/presentation.py
+++ b/coda/coda_replication/presentation.py
@@ -33,32 +33,35 @@ def queue_object_to_xml(queue_entry, webRoot=None):
     return atomXML
 
 
-def xmlToQueueEntry(queueXML):
-    """
-    Given some XML, create a Queue entry with it
-    """
+def xmlToQueueEntry(queue_xml):
+    """Convert XML to a QueueEntry object."""
 
-    queueEntryNode = queueXML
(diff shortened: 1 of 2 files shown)


damonkelley:
ping @somexpert  @ldko

somexpert:
:thumbsup:

ldko:
Of partial relevance to this PR, I noticed xmlToQueueEntry is unnecessarily imported by views.py.

damonkelley:
That is probably best left for a PR where the views imports need to be touched. There were a few  imports in `presentation.py` that were able to be removed, so went ahead and cleaned up all of them.

ldko:
:+1:

PR #80, opened 2016-01-28
Remove duplicated functions.

damonkelley:
`getValueByName` and `getNodeByName` are functions with equivalent
implementations that are already available within the codalib library.


damonkelley:
ping @somexpert @ldko

somexpert:
:thumbsup:

ldko:
:+1:

PR #83, opened 2016-02-01
MDStore APP Datetime Format

damonkelley:
ping @ldko @somexpert


ldko:
It's unfortunate that the individual node entries are created with `nodeEntry` while the node feed is made up of entries generated by `codalib.bagatom.nodeToXML` so the node feed still has entries with the old date format. There is unt-libraries/codalib#7 though.

damonkelley:
I wonder if any codalib clients use that function (or any of the other related functions in the module). If not it would be more appropriate for them to be a part of this project.

ldko:
Looking just now I didn't see other calls of `codalib.bagatom.nodeToXML` in our internal coda clients. Just the one call in `coda_mdstore/views.py`. However, `codalib.bagatom.queueEntryToXML` is similarly used to make both the feed and individual queue entry XML in the `coda_replication` app and _is_ also used by internal clients.

damonkelley:
ok, well we can move that over then. 

FWIW, at some point in the near future we are going to need to reevaluate our approach XML serialization and deserialization. I am confident we can rethink it and come up with a more modular and more maintainable approach.

damonkelley:
This has uncovered that we have multiple functions generating node xml, and that the separate implementations are generating **different** xml, but both are using the same incorrect datetime 

Going to close this for now, because we are going to consolidate the xml generation in the process.

PR #85, opened 2016-02-08
Premis templates inherit from the project's base template.

damonkelley:
This gives the premis app the same look and feel as the rest of the project.


damonkelley:
ping @ldko @somexpert

somexpert:
:thumbsup:

ldko:
It looks like the premis event views had been changed on our internal code base after premis_event_service went on to GitHub to send `site_title` to templates and thus match the other coda apps. Currently using the coda_mdstore base template, the templates for the premis app in GitHub won't be getting a `site_title` so the site name is missing in the brand/logo at the top left corner of the premis templates in this project. Would it work to do something like write a context processor to add to  `TEMPLATE_CONTEXT_PROCESSORS` so we can get the site name in the templates?

damonkelley:
Yes

damonkelley:
In fact that is preferable to the way the current coda apps are doing it, which is to add `site_title` to every views template context, which is a pretty significant DRY violation.

damonkelley:
@ldko, is that something you would like taken care of now? It seems like not having the `site_title` for the premis event app wont be much of an issue. Only cosmetic.

ldko:
I can open an issue for later, and you can merge this now. :+1:

damonkelley:
Sounds good. I really think the whole `site_title` might need to be re-examined anyway. We are doing a db query on for that on every request. Seems like something more suited for a setting.

PR #91, opened 2016-04-15
Add /validate/check.json

### LINKED ISSUE #90: Add view of validate `last_verified_status` counts in JSON format
In coda_validate, add a view, perhaps at /validate/check/, that returns JSON formatted data containing the number of entries for each of the Passed/Failed/Unverified values of `last_verified_status`.

Like issue #10 and #70, this will improve visibility of failed validations. This view will be useful for consumption by non-humans.

damonkelley:
Closes #90

### DIFF
--- a/coda/coda_validate/models.py
+++ b/coda/coda_validate/models.py
@@ -1,7 +1,30 @@
 from django.db import models
 from datetime import datetime
 
-# Create your models here.
+
+class ValidateManager(models.Manager):
+
+    def last_verified_status_counts(self):
+        results = self._verified_counts()
+        results = self._format_count_results(results)
+        return self._set_default_status_counts(results)
+
+    def _set_default_status_counts(self, counts):
+        for _, value in Validate.VERIFIED_STATUS_CHOICES:
+            counts.setdefault(value, 0)
+        return counts
+
+    def _format_count_results(self, results):
+        return {
+            c['last_verified_status']: c['count']
+            for c in results
+        }
+
+    def _verified_counts(self):
+        return (self.values('last_verified_status')
+                    .annotate(count=models.Count('last_verified_status')))
+
+
 class Validate(models.Model):
     VERIFIED_STATUS_CHOICES = (
         ("Unverified", "Unverified"),
@@ -10,13 +33,15 @@ class Validate(models.Model):
     )
     identifier = models.CharField(max_length=255, unique=True, editable=False)
     added = models.DateTimeField(auto_now_add=True)
-    last_verified = models.DateTimeField(default=datetime(2000,01,01))
+    last_verified = models.DateTimeField(default=datetime(2000, 1, 1))
     last_verified_status = models.CharField(
         max_length=25, choices=VERIFIED_STATUS_CHOICES, default="Unverified"
     )
-    priority_change_date = models.DateTimeField(default=datetime(2000,01,01))
+    priority_change_date = models.DateTimeField(default=datetime(2000, 1, 1))
     priority = models.IntegerField(default=0)
     server = models.CharField(max_length=255)
+
+    objects = ValidateManager()
 
     def __unicode__(self):
         return '%s' % self.identifier
--- a/coda/coda_validate/models.py
+++ b/coda/coda_validate/models.py
@@ -6,23 +6,29 @@ class ValidateManager(models.Manager):
 
     def last_verified_status_counts(self):
         results = self._verified_counts()
-        results = self._format_count_results(results)
-        return self._set_default_status_counts(results)
-
-    def _set_default_status_counts(self, counts):
-        for _, value in Validate.VERIFIED_STATUS_CHOICES:
-            counts.setdefault(value, 0)
-        return counts
-
-    def _format_count_results(self, results):
-        return {
-            c['last_verified_status']: c['count']
-            for c in results
-        }
+        return VerifiedCountsResultFormatter(results).format()
 
     def _verified_counts(self):
         return (self.values('last_verified_status')
                     .annotate(count=models.Count('last_verified_status')))
+
+
+class VerifiedCountsResultFormatter(object):
+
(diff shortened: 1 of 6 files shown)


damonkelley:
ping @ldko @somexpert

somexpert:
I don't necessarily _disagree_ with this choice, but I'm curious as to why you decided to include the status code assertion with the count assertions in your tests.

somexpert:
:thumbsup:

coda/coda_validate/urls.py
```diff
@@ -6,6 +6,7 @@
     url(r'^APP/validate/(?P<identifier>.+?)/$', views.app_validate, name='app-detail'),
     url(r'^APP/validate/$', views.app_validate, name='app-list'),
     url(r'^validate/$', views.index, name='index'),
+    url(r'^validate/check.json/$', views.check_json, name='check_json'),
```
ldko: Should we lose the end slash on the check.json URL?

coda/coda_validate/tests/test_views.py
```diff
@@ -369,3 +369,27 @@ def test_invalid_method_and_identifier_returns_bad_request(self, rf):
...
+    def test_with_statuses(self, rf):
+        factories.ValidateFactory.create_batch(100)
+        request = rf.get('/')
+        response = views.check_json(request)
+        assert response.status_code == 200
+
+        counts = json.loads(response.content)
+        assert counts['Passed'] > 0
+        assert counts['Failed'] > 0
+        assert counts['Unverified'] > 0
```
ldko: When we are asserting that the counts of the different statuses are greater than 0 here, are we basing this passing on a strong likelihood that the ValidateFactory will create validates with every status in a relatively large batch or is it guaranteed that ValidateFactory will create validate entries with each of the statuses?

coda/coda_validate/urls.py
ldko: We could add a test for the new url.

coda/coda_validate/tests/test_views.py
damonkelley: No, there is the very small possibility that a `Validate` object of with one of the three statuses will not be created. Assuming my math is correct the chance of that scenario occurring given is `1.9403252e-48` :smile:

coda/coda_validate/urls.py
damonkelley: I am not sure if there should be a slash or not. I personally think that there should be no trailing slashes on any of the URLs, but its not my decision. Tell me what you and @vphill want, and I will make it so :picard_meme:

:smile_cat:

ldko: Let's lose the end slash after the check.json.

coda/coda_validate/tests/test_views.py
ldko: Since you could guarantee the results with at most 1 extra line of code, I would rather not get into the habit of us writing tests that allow for possible known failure.

damonkelley: what extra line is that?

ldko:
:+1:

coda/coda_validate/models.py
```diff
@@ -1,7 +1,36 @@
...
+
+class ValidateManager(models.Manager):
+
+    def last_verified_status_counts(self):
+        results = self._verified_counts()
+        return VerifiedCountsResultFormatter(results).format()
+
+    def _verified_counts(self):
+        return (self.values('last_verified_status')
+                    .annotate(count=models.Count('last_verified_status')))
```
ldko: In running coda to look at #95, I am noticing unexpected results being returned from this aggregation.
```
[{'count': 1, 'last_verified_status': u'Unverified'}, {'count': 1, 'last_verified_status': u'Unverified'}, {'count': 1, 'last_verified_status': u'Passed'}, {'count': 1, 'last_verified_status': u'Unverified'}, {'count': 1, 'last_verified_status': u'Failed'}, {'count': 1, 'last_verified_status': u'Failed'}]
```
Explicitly getting rid of the query ordering using `order_by()` is fixing the issue for me:
```
         return (self.values('last_verified_status')
                    .annotate(count=models.Count('last_verified_status')).order_by())
```
Has anyone else noticed this behavior?

damonkelley: No. But I have similar issues with in other apps. You might try rebuilding the app container.

ldko: I have.

damonkelley: Was this from running the tests, or did you discover this when you went to `/validate/check.json`?

Also, did you remove the existing container before you started the freshly built image?

ldko: I noticed it on `/validate/` and then `/validate/check.json`.
I believe I stopped the containers, removed the app container, and then built again.

ldko: Can I add order_by() to this, or would it be problematic, or are you thinking it is an unnecessary addition?

damonkelley: I am hesitant because this isn't happening in the Travis build, or in my local environment.

damonkelley: Do all of the tests pass when you run them locally?

ldko: The tests pass. I am wondering though if the Validate Factory could be generating validate objects with the same `added` date which could prevent the issue I'm having to show if the issue is related to what is here in the documentation https://docs.djangoproject.com/en/1.8/topics/db/aggregation/#interaction-with-default-ordering-or-order-by. That documentation is what is leading me to want to add `order_by()`.

damonkelley: That is a good point. Lets see if we can get the tests to fail first before make any changes.

damonkelley: Ok I see it now.

What makes this hard to test is that that it uses orders by the `added` field by default, and that field is the auto added, which makes it hard to provide a fake timestamp.

PR #93, opened 2016-04-25
Validate last status count labels

### LINKED ISSUE #10: coda_validate status buttons
@vphill would like to see green, red, and yellow buttons (similar to the ones on the index page of coda_mdstore but smaller) for the different validate statuses that are Passed, Failed, and Unverified. These might link to the entries that have the specified status. These buttons should be placed on the coda_validate index page at `/validate/` under the Coda Validate logo and above the "Prioritized Items" section.

damonkelley:
Fixes #10 

### With validate objects
selection_015

### Without validate objects
selection_014

ping @ldko @vphill @somexpert

### DIFF
--- a/coda/templates/coda_validate/index.html
+++ b/coda/templates/coda_validate/index.html
@@ -6,6 +6,13 @@
 <div class="well pagination-centered">
     <h1>Coda <i class="icon-ok"></i>alidate</h1>
 </div>
+
+<ul class="well pull-right inline">
+    <li>Passed <span class="badge badge-success">{{ verified_counts.Passed }}</span></li>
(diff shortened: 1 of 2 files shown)


ldko:
I like the way this looks. :+1:

coda/templates/coda_validate/index.html
```diff
@@ -6,6 +6,13 @@
 <div class="well pagination-centered">
     <h1>Coda <i class="icon-ok"></i>alidate</h1>
 </div>
+
+<ul class="well pull-right inline">
+    <li>Passed <span class="badge badge-success"> {{ verified_counts.Passed }}</span></li>
```
somexpert: Got a space between `>` and `{{` we can get rid of. Other than that, lgtm!

PR #96, opened 2016-04-29
Remove ordering for counting validate statuses

ldko:
This should fix an issue of `Validate`'s default ordering by the `added` field causing unexpected results on getting the `last_verified_status` counts for objects.


damonkelley:
:+1:

somexpert:
:thumbsup:

PR #97, opened 2016-05-10
Display all nodes on one page for APP feed.

ldko:
We want to display all node entries at `/APP/node` on one page.

@vphill is this kind of what you are thinking?

### DIFF
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -935,7 +935,8 @@ def app_node(request, identifier=None):
         return resp
     # FEED
     elif request.method == 'GET' and not identifier:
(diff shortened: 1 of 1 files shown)


vphill:
:+1:

ldko:
@somexpert @damonkelley here is a one line change for review.

somexpert:
:thumbsup:

damonkelley:
As mentioned in Slack, we can reduce this to one query with something like the following.

```python
nodes = Node.objects.all()
paginator = Paginator(nodes, len(nodes))
# ...
```

damonkelley:
:+1:

Issue #92, opened 2016-04-19
Fail on missing template variable


damonkelley:
Add the `--fail-on-template-vars` option in the pytest settings. This will cause a test failure any time a view renders a template that attempts to use a variable that is not defined in the context.

runderwood:
Closed via #101

Issue #86, opened 2016-02-09
Pass `site` to templates via context processor


ldko:
Currently using the coda_mdstore base template, the templates for the premis app in this project aren't  getting a `site_title` so the site name is missing in the brand/logo at the top left corner of the premis pages. We could write a context processor that we add to `TEMPLATE_CONTEXT_PROCESSORS` to make the site name available in all the templates (and remove the many places where we are passing `site_title` via views).

This issue came up in discussion about PR #85.

damonkelley:
Currently, we are getting the `site_title` by querying the database for this value. We should consider moving this to the settings file instead.

runderwood:
Closed via #101

Issue #68, opened 2016-01-07
Update Template Settings


damonkelley:
The pre-1.8 template settings have been deprecated. Update the settings to use the new `TEMPLATE` settings.

runderwood:
Closed via #101

PR #101, opened 2017-01-23
Smallfixes

runderwood:
@ldko @somexpert refs. #68, #86, and #92

### DIFF
--- a/coda/coda_mdstore/context.py
+++ b/coda/coda_mdstore/context.py
@@ -1,0 +1,9 @@
+from django.contrib.sites.models import Site
+
+def site_info(request,):
+    # stash site title on first call
+    try:
+        sinfo = site_info.cached_site_info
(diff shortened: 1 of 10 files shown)


ldko:
Regarding updating [template settings to 1.8](https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/), I think we could further comply by:
* Replacing `django.core.context_processors` with `django.template.context_processors` in the names of the context processors
* Remove `TEMPLATE_DEBUG` setting since it is same as DEBUG

runderwood:
Removed references to `TEMPLATE_DEBUG` and `django.core.context_processors`. Replaced the latter w/ `django.template.context_processors` in 97f50b0.

somexpert:
Looks good to me! :thumbsup:

ldko:
me too :+1:

PR #103, opened 2017-03-15
Atom related fixes, PREMIS validation, PEP8 conformity, Docker config

runderwood:
@ldko @somexpert I'm very sorry about this -- it is way too big. I did it when I first started, and it kind of kept growing. A lot of it is just PEP8 stuff to pass flake8 tests. But there are a number of fixes to the AtomPub facilities. There are also some changes to the Docker mechanics and settings, and the requirements for `codalib` and `django-premis-event-service` have been synch'd with upstream changes.


coda/coda_replication/tests/test_views.py
```diff
@@ -384,7 +385,8 @@ def test_get_with_invalid_identifier_returns_not_found(self, rf):
 
     def test_delete(self, rf):
         entry = factories.QueueEntryFactory.create()
-        request = rf.delete('/')
+        print entry.ark
+        request = rf.delete('/'+entry.ark)
         response = views.queue(request, entry.ark)
```
ldko: Do we need these changes?

coda/coda_validate/views.py
```diff
@@ -172,7 +174,8 @@ def stats(request):
     # make a set of data that makes sense for the heatmap
     all_v = Validate.objects.all().order_by('last_verified')
     sums_by_date = {}
-    first_day, last_day = all_v[0], all_v.reverse()[0]
+    last_day = all_v[0], all_v.reverse()[0]
+    last_day = last_day[-1]
```
ldko: These two lines seem strange. Should they be replaced with something like `last_day = all_v[-1]`?

runderwood: This is ugly, but it's because you can't negative index this queryset. The second call is after the queryset evaluation has been triggered. Pushing a cleaner version of this. The two values don't need to be there at all -- it's a holdover from when an unused value was also set here.

coda/coda_validate/views.py
```diff
@@ -77,8 +77,8 @@ def items(self, obj):
             # instead, let's do this:
             # http://elpenia.wordpress.com/2010/05/11/getting-random-objects-from-a-queryset-in-django/
             v = validations.filter(
-                last_verified__lte=datetime.datetime.now() - \
-                    settings.VALIDATION_PERIOD
+                last_verified__lte=datetime.datetime.now() -
```
somexpert: So I'm inclined to agree with [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) on their recommendation to put operators with their operands on the new line so it is easier to tell what is being subtracted in this case. I am not saying _this_ should be changed, but just wanted to throw it out there as a possible suggestion. I haven't been doing this myself but I think I will start to.

somexpert:
:thumbsup:

ldko:
Looks good :+1:

PR #106, opened 2017-03-20
Clean up mdstore_oai, closes #27

### LINKED ISSUE #27: Clean up md_storeOAIInterface
There is a lot we can do to clean up `code_mdstore.mdstore_oai.mdstore_OAIInterface`. The docstring states that it implements the `ResumptionOAIPHM` interface, but it more accurately implements the `IOAI` interface [here](https://github.com/infrae/pyoai/blob/master/src/oaipmh/interfaces.py)

For starters
- Remove `_get_old_solr_getRecord`
- Remove `setAssociations`
- At the very least, rename `listStuff` to something more descriptive. It is more likely that we can remove `listStuff` and just implement the functionality into the two methods that call it.

runderwood:
Ticks remaining boxes in #27

### DIFF
--- a/coda/coda_mdstore/mdstore_oai.py
+++ b/coda/coda_mdstore/mdstore_oai.py
@@ -143,8 +143,7 @@ class OAIInterface(object):
 def makeDataRecord(
     bagObject,
     domain="",
-    metadataPrefix="oai_dc",
-    setAssociationList=[]
+    metadataPrefix="oai_dc"
 ):
     """
     Given a Bag object, make a record in the format that oaipmh uses
--- a/coda/coda_mdstore/mdstore_oai.py
+++ b/coda/coda_mdstore/mdstore_oai.py
@@ -23,11 +23,6 @@ class OAIInterface(object):
             # 'granularity':'YYYY-MM-DD',
             'compression': ['identity'],
         }
-        self.setAssociations = [
-            ("dc_rights_access", "access_rights"),
-            ("untl_institution", "partner"),
-            ("untl_collection", "collection"),
-        ]
         if identifyDict:
             self.identifyDict.update(identifyDict)
         self.resultSize = resultSize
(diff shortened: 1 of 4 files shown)


coda/coda_mdstore/mdstore_oai.py
```diff
@@ -95,14 +94,14 @@ def listIdentifiers(self, metadataPrefix=None, from_=None, until=None,
...
-        return self.listStuff(
+        return self.makeList(
             metadataPrefix=metadataPrefix, from_=from_,
             until=until, set=set, cursor=cursor, batch_size=batch_size,
             headersOnly=False
         )
 
-    def listStuff(self, metadataPrefix=None, from_=None, until=None,
-                  set=None, cursor=0, batch_size=10, headersOnly=True):
+    def makeList(self, metadataPrefix=None, from_=None, until=None,
```
somexpert: I think I'd personally like to see a bit more descriptive name, or instead, a more descriptive docstring.

ldko: I agree; if we are changing the name from `listStuff`, more description would be worthwhile.

runderwood: I'm not sure how to name the method more descriptively. It's within the context of an OAI interface, so I feel like the sort of generic list it refers to within that context would be fairly discernible to a hypothetical reader. I can add a note about making generic lists.

somexpert: Well, it gives more info than was previously there, so works for me. :thumbsup:

ldko:
Do we need to keep `setAssociationList` here?
```
def makeDataRecord(
    bagObject,
    domain="",
    metadataPrefix="oai_dc",
    setAssociationList=[]
):
```

runderwood:
@ldko No! Good catch.

PR #107, opened 2017-03-23
Fixes for validate statistics

### LINKED ISSUE #9: Cleanup coda_validate statistics
On large scale data (we are at just under 1.3 million validate entries), the statistics run very slowly. We should reduce the number of queries used to generate the stats, perhaps by adding a model manager similar to what Damon did in the [name app](https://github.com/unt-libraries/django-name). Also, the stats numbers shown in the two visualizations do not seem to match the text statistics where they should. The "Unverified" in the pie chart doesn't match "Unverified in Queue." The "Passed" in the pie chart doesn't match "Past 365 days" though all validations have been run this year and all passed. The number in the calendar visualization for July 1 do not match the number in "This month."

coda_validate_unverified_2015-07-01

coda_validate_passed_2015-07-01

coda_validate_month_2015-07-01

runderwood:
@ldko @somexpert Should close #9. Fixes slow looping code and redundant queries by using raw sql for aggregates. Also fixes obscuring of multiple calendars by setting the Google chart height dynamically.
2017-03-23-091415_1920x1080_scrot

### DIFF
--- a/coda/coda_validate/models.py
+++ b/coda/coda_validate/models.py
@@ -75,18 +75,3 @@ class Validate(models.Model):
                 sums_by_date[dt] = ct
                 row = cur.fetchone()
         return sums_by_date
-
-    @classmethod
-    def result_counts(cls,):
-        sql = """select last_verified_status as st, count(*) as ct
-        from %s
-        group by last_verified_status""" % (cls._meta.db_table,)
-        results = {}
-        with connection.cursor() as cur:
-            cur.execute(sql)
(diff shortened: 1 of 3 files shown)


ldko:
Do we need to use raw SQL here? I thought Django ORM has aggregate functionality available.

runderwood:
Doing the date conversion doesn't appear to be doable with an aggregate in the ORM. Likewise, I cannot figure out how to do mutliple aggregate counts with any precision.

Issue #110, opened 2017-06-30
Incorrectly using of link_object_id instead of linked_object_id


ldko:
We are using `link_object_id` when it should be `linked_object_id` on the [bag info page](https://github.com/unt-libraries/coda/blob/363125c8623a50208b3d9d0615ba1e640a35a2d5/coda/coda_mdstore/views.py#L498) and on the [about page](https://github.com/unt-libraries/coda/blob/363125c8623a50208b3d9d0615ba1e640a35a2d5/coda/templates/mdstore/about.html#L50). This is causing a bag to show events that are not associated with it because the search returns results for any linked object.

Another thing, where the template says ["There are...premis events associated with..."](https://github.com/unt-libraries/coda/blob/363125c8623a50208b3d9d0615ba1e640a35a2d5/coda/templates/mdstore/bag_info.html#L49) it is not going to say higher than 20, since the max number of events given by search.json on a page (which determines the count shown) is 20. Should we change the language to reflect that there are up to 20 events shown, but there may be more? <-- @vphill  (We link to the event search, so it would be easy to get to the rest.)

Which brings up another thing because the page that you get linked to for seeing all the events for the bag, `event/search/?linked_object_id=someid`, has pagination, but the pagination links get an empty `linked_object_id` value (such as  `event/search/?page=2&event_outcome=&event_type=&start_date=&end_date=&linked_object_id=&min_ordinal=204373`) so you get more unassociated results on those pages.

ldko:
Also, search.json throws a 500 error if you pass a `linked_object_id` that does not exist.

PR #111, opened 2017-07-05
Fix 110

ldko:
This replaces incorrect use of `link_object_id` with `linked_object_id` and tweaks the language of the `bag_info.html` template to reflect the limited number of PREMIS events that will be shown.

These changes assume we are ok with only showing the top 20 results of events on the bag info page as is brought up in #110 . If we want to show them all, we can address that...

@runderwood @vphill do you want to look at this? There are still a couple of things that need looking at in PES related to searching bags with `linked_object_id` for which I created an issue on [PES](https://github.com/unt-libraries/django-premis-event-service/issues/82).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.swo
 
 secrets.json
+env/
 .coverage
 htmlcov/*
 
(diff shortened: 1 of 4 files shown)


runderwood:
:+1:

I'll handle paging!

PR #112, opened 2017-07-06
Include all linked events in bag info

### LINKED ISSUE #110: Incorrectly using of link_object_id instead of linked_object_id

runderwood:
Should close #110. (Also refs #111).

@ldko @somexpert

### DIFF
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -483,23 +483,16 @@ def bagHTML(request, identifier):
     # try and get the bag
     bag = get_object_or_404(Bag, name__exact=identifier)
     bag_info = Bag_Info.objects.filter(bag_name__exact=bag)
-    # determine some values that will be databased better in the future
+    # Put the bag info in a dict
+    bag_info_d = dict((info.field_name, info.field_body) for info in bag_info)
+    oxum_count = oxum_size = -1
     try:
-        payload_oxum_file_count = (
-            item for item in bag_info if item.field_name == "Payload-Oxum"
-        ).next().field_body.split(".")[1]
+        oxum = bag_info_d.get('Payload-Oxum', '')
+        oxum_count, oxum_size = map(int, oxum.split('.', 1))
     except:
-        payload_oxum_file_count = -1
+        pass
     try:
-        payload_oxum_size = (
-            item for item in bag_info if item.field_name == "Payload-Oxum"
-        ).next().field_body.split(".")[0]
-    except:
-        payload_oxum_size = -1
-    try:
-        bag_date = datetime.strptime((
-            item for item in bag_info if item.field_name == "Bagging-Date"
-        ).next().field_body, '%Y-%m-%d')
+        bag_date = datetime.strptime(bag_info_d.get('Bagging-Date'), '%Y-%m-%d')
     except:
         bag_date = None
     # grab the related premis events from the json search for
@@ -518,6 +511,7 @@ def bagHTML(request, identifier):
         if json_events:
             total_events = json_events.get('feed', {})
             total_events = total_events.get('opensearch:totalResults', 0)
+        # Follow next url in results until we hit the last page
         while json_events:
             this_page = json_events.get('feed', {})
             this_page = this_page.get('entry', [])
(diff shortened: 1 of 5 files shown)


coda/coda_mdstore/views.py
```diff
@@ -483,29 +484,62 @@ def bagHTML(request, identifier):
...
+            item for item in bag_info if item.field_name == "Payload-Oxum"
+        ).next().field_body.split(".")[1]
+    except:
+        payload_oxum_file_count = -1
+    try:
+        payload_oxum_size = (
+            item for item in bag_info if item.field_name == "Payload-Oxum"
+        ).next().field_body.split(".")[0]
+    except:
+        payload_oxum_size = -1
```
ldko: Is it better for the `except` statements to return a string for `payload_oxum_file_count` and `payload_oxum_size`? Would it be better to write `487-498` in a way that only iterates the `bag_info` items looking for `Payload-Oxum` once?

coda/coda_mdstore/views.py
```diff
@@ -483,29 +484,62 @@ def bagHTML(request, identifier):
...
+    try:
+        bag_date = datetime.strptime((
+            item for item in bag_info if item.field_name == "Bagging-Date"
+        ).next().field_body, '%Y-%m-%d')
+    except:
+        bag_date = None
```
ldko: Does the template handle a `None` value cleanly at `<i class="icon-calendar"></i> {{ bag_date|timesince }} ago`? I don't know if it does or not and haven't checked.

coda/coda_mdstore/views.py
```diff
@@ -483,29 +484,62 @@ def bagHTML(request, identifier):
...
+                try:
+                    json_response = urlopen(next_url)
+                    json_events = json.load(json_response)
+                except:
+                    json_events = None
+            else:
+                json_events = None
     except:
-        json_events = json.loads('{}')
+        json_events = None
```
ldko: Does setting `json_events` here do anything, since it is no longer sent to the template?

coda/coda_mdstore/views.py
```diff
@@ -483,29 +484,62 @@ def bagHTML(request, identifier):
...
     return render_to_response(
         'mdstore/bag_info.html',
         {
-            'json_events': json_events,
+            'linked_events': linked_events, 'total_events': total_events,
```
ldko: me nitpicking :grin: do you mind putting `linked_events` and `total_events` on separate lines?

coda/coda_mdstore/views.py
runderwood: It would be better to only iterate once. I just went with what was there. But I'll tighten it up.

I set them as negative integers to distinguish them, but I don't suppose it matters much since I didn't update the templates accordingly.

coda/coda_mdstore/views.py
runderwood: I intended to update the template but didn't. It should differentiate between no-date and yes-we-have-date.

coda/coda_mdstore/views.py
runderwood: Well, I didn't set it outside of the try block, so I typically like to make sure the variable is set either way.

I also intended to use it in the template but didn't in the end.

coda/coda_mdstore/views.py
runderwood: Actually, it doesn't crater anything, because the template checks for the 'Bagging-Date' bag-info field.

coda/coda_mdstore/views.py
runderwood: Fixed in 2c7a61b.

somexpert:
LGTM :thumbsup:

ldko:
it works for me! :+1:

PR #119, opened 2018-11-30
Proxy downloads

ldko:
This PR is intended to allow us to serve a bag's data files (via `bagProxy`) by having our proxy server get the data files from directly where they are hosted rather than pull them through Django. This should fix an issue we have when users try to download some larger files.

ping @somexpert @vphill

### DIFF
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -644,7 +644,7 @@ def bagProxy(request, identifier, filePath):
             content_type=handle.info().getheader('Content-Type')
         )
         resp['Content-Length'] = handle.info().getheader('Content-Length')
-        if hasattr(settings, 'REPROXY') and settings.REPROXY:
+        if getattr(settings, 'REPROXY', False):
             # Have a proxy server point the client to where to download
             # the file directly in order to bypass serving through Django.
             resp['X-REPROXY-URL'] = handle.geturl()
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -1,5 +1,7 @@
 import codecs
 import copy
+import uuid
+
 from urllib2 import urlopen
 from urllib import urlencode
 try:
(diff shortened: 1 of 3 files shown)


coda/coda_mdstore/views.py
```diff
@@ -639,10 +641,17 @@ def bagProxy(request, identifier, filePath):
     handle = getFileHandle(identifier, filePath)
     if handle:
         resp = HttpResponse(
-            FileWrapper(handle),
             content_type=handle.info().getheader('Content-Type')
         )
         resp['Content-Length'] = handle.info().getheader('Content-Length')
+        if hasattr(settings, 'REPROXY') and settings.REPROXY:
```
somexpert: What if we use getattr to do both these checks?
```suggestion
        if getattr(settings, 'REPROXY', False):
```

coda/coda_mdstore/views.py
```diff
@@ -639,10 +641,17 @@ def bagProxy(request, identifier, filePath):
...
+            # Have a proxy server point the client to where to download
+            # the file directly in order to bypass serving through Django.
+            resp['X-REPROXY-URL'] = handle.geturl()
+            resp['ETag'] = '"%s"' % uuid.uuid4().get_hex()
```
somexpert: I'm a little confused about the use of quotes here.

ldko: `ETag` header values are specified to be in [double quotes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#Directives). Is that what you are wondering about?

somexpert: yep, that was it. thanks for clearing that up!

somexpert:
How do you feel about a test that checks for the appropriate headers in the response based on the presence of the REPROXY setting? I'm not sure how I myself feel about whether it's worth it or not...

PR #120, opened 2019-03-01
Fix pytest version, fix flake8 error, and add pytest-flake8.

somexpert:
ping @ldko, @madhulika95b, this is ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,11 @@ $ docker-compose up -d app
 
 ## Running the tests
 
+The db container must already be running, or the tests will probably
+fail on the first run since the database needs time to warm up.
+
 ```sh
-# On the initial run or if the schema changes.
-# The db container must already be running, or the tests will probably
-# fail on the first run since the database needs time to warm up.
+# On the initial run or if the schema changes
 $ docker-compose run --rm test --create-db
 
 # Subsequent runs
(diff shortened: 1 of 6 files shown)


somexpert:
Yep, I had deleted the db container and decided to see how this would work without having that container warmed up, and there were a ton of failures, and from what I saw they were all related to the DB connection.

README.md
```diff
@@ -60,7 +62,9 @@ $ docker-compose up -d app
 ## Running the tests
 
 ```sh
-# On the initial run or if the schema changes
+# On the initial run or if the schema changes.
+# The db container must already be running, or the tests will probably
+# fail on the first run since the database needs time to warm up.
```
ldko: Having the explanation in these comments as 2 sentences (one that isn't really a sentence) reads a little awkwardly to me. I would probably move the explanatory text outside of comments, but if you prefer to leave them here, I am ok with that.

requirements.txt
```diff
@@ -6,8 +6,10 @@ MySQL-python==1.2.3rc1
 python-dateutil==1.4.1
```
ldko: I think it makes sense to also bump some other versions:
django==1.8.19
lxml==3.3.0
python-dateutil==2.4

README.md
somexpert: done....

requirements.txt
somexpert: .....and done

PR #126, opened 2019-08-29
Django19

madhulika95b:
@somexpert @ldko Please review the upgrade to django 1.9


coda/coda_mdstore/views.py
```diff
@@ -662,14 +662,14 @@ def externalIdentifierSearch(request, identifier=None):
...
-                request.REQUEST.get('ark'):
-            identifier = 'ark:/%d/%s' % (settings.ARK_NAAN, request.REQUEST.get('ark'))
+        if request.POST.get('ark') and ('ark:/%d' % settings.ARK_NAAN) not in \
+                request.POST.get('ark'):
+            identifier = 'ark:/%d/%s' % (settings.ARK_NAAN, request.POST.get('ark'))
             feed_id = request.path + identifier + '/'
-        elif request.REQUEST.get('ark'):
-            identifier = request.REQUEST.get('ark')
+        elif request.POST.get('ark'):
+            identifier = request.POST.get('ark')
```
ldko: It looks to me like we should actually be using `request.GET.get` in these changes, not `request.POST.get`. This is based on:
1. in the test for this view, `request = rf.get(url, {'ark': 'ark:/%d/metadc000001' % (settings.ARK_NAAN,)})` using a `get` method
2. having to change the call in the test to get it to pass with this change
3. Looking at the form at `/extidentifier/` which shows `get` method `<form action="/extidentifier/" method="get">`

coda/coda_oaipmh/views.py
```diff
@@ -29,5 +29,5 @@ def index(request):
         resumption_batch_size=500
     )
 
-    response = oaiServer.handleRequest(request.REQUEST)
+    response = oaiServer.handleRequest(request.POST)
```
ldko: I suspect this should also be a `request.GET` not `request.POST`. To help with checking this, I think it would be good to [test this view](https://github.com/unt-libraries/coda/blob/acbd6fdb95c208b584fe6a468472f3308b50fda6/coda/coda_oaipmh/tests/test_views.py#L9) more than just test the content_type and response code.

I notice that http://localhost:8787/oai/?verb=Identify when using `request.POST` we get XML and 200, but the response is:
```
?xml version='1.0' encoding='UTF-8'?>
<OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
  <responseDate>2019-08-29T22:38:03Z</responseDate>
  <request>http://example.com/oai/</request>
  <error code="badVerb">Required verb argument not found.</error>
</OAI-PMH>
```
with a `request.GET` the response is again XML and 200, but the content is more like:
```
<?xml version='1.0' encoding='UTF-8'?>
<OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
  <responseDate>2019-08-29T22:41:55Z</responseDate>
  <request verb="Identify">http://example.com/oai/</request>
  <Identify>
    <repositoryName>example.com</repositoryName>
    <baseURL>http://example.com/oai/</baseURL>
    <protocolVersion>2.0</protocolVersion>
    <adminEmail>mark.phillips@unt.edu</adminEmail>
    <earliestDatestamp>2004-05-19T00:00:00Z</earliestDatestamp>
    <deletedRecord>transient</deletedRecord>
    <granularity>YYYY-MM-DDThh:mm:ssZ</granularity>
    <description>
      <toolkit xmlns="http://oai.dlib.vt.edu/OAI/metadata/toolkit" xsi:schemaLocation="http://oai.dlib.vt.edu/OAI/metadata/toolkit http://oai.dlib.vt.edu/OAI/metadata/toolkit.xsd">
        <title>pyoai</title>
        <version>2.4.4</version>
        <URL>http://infrae.com/products/oaipack</URL>
      </toolkit>
    </description>
  </Identify>
</OAI-PMH>
```

ldko:
I notice there is reference to `request.REQUEST` in `coda_mdstore.views.bagFullTextSearchAtom` that would need to be changed, but looking at the code and the test for it, it looks like we don't use this view. Perhaps @vphill could tell us if that is something we want to keep and use (and we would need to change it for Django 1.11) or if we could delete it.

coda/coda_mdstore/views.py
madhulika95b: ok! I'll try using GET instead of POST here and undo the changes in tests and see if it works.

ldko:
> I notice there is reference to `request.REQUEST` in `coda_mdstore.views.bagFullTextSearchAtom` that would need to be changed, but looking at the code and the test for it, it looks like we don't use this view. Perhaps @vphill could tell us if that is something we want to keep and use (and we would need to change it for Django 1.11) or if we could delete it.

Since we haven't heard back about this, perhaps go ahead and upgrade it and add a functioning test.  I will open an issue about what to do with it in the future.

ldko:
Nevermind above, we are going to remove `bagFullTextSearchATOM` and accompanying stub test.

PR #129, opened 2019-09-10
Django110

madhulika95b:
@ldko @somexpert Upgraded to django 1.10 version. Please review it.

### DIFF
--- a/coda/coda_mdstore/urls.py
+++ b/coda/coda_mdstore/urls.py
@@ -1,6 +1,7 @@
-from django.conf.urls import url, patterns
+from django.conf.urls import url
 from . import views
-from .resourcesync import sitemaps
+from django.contrib.sitemaps import views as sitemap_views
+from .resourcesync import changelist, capabilitylist
 
 urlpatterns = [
     url(r'^bag/$', views.all_bags, name='bag-list'),
@@ -31,42 +32,9 @@ urlpatterns = [
     url(r'^about/$', views.about, name='about'),
     url(r'^robots.txt$', views.shooRobot, name='robots'),
     url(r'^feed/$', views.AtomSiteNewsFeed(), name='feed'),
+    url(r'^resourceindex\.xml$', sitemap_views.index),
+    url(r'^resourcelist-(?P<section>.+)\.xml$', sitemap_views.sitemap, name='sitemaps_views'),
+    url(r'changelist\.xml$', changelist),
+    url(r'^capabilitylist\.xml$', capabilitylist),
     url(r'^$', views.index, name='index'),
 ]
-
-
-urlpatterns += patterns(
-    'django.contrib.sitemaps.views',
-    (
-        r'^resourceindex\.xml$',
-        'index',
-        {
-            'sitemaps': sitemaps,
-            'template_name': 'mdstore/resourceindex.xml'
-        },
-    ),
-    (
-        r'^resourcelist-(?P<section>.+)\.xml$',
-        'sitemap',
-        {
-            'sitemaps': sitemaps,
-            'template_name': 'mdstore/sitemap.xml'
-        },
-    ),
-)
-urlpatterns += patterns(
-    'coda_mdstore.resourcesync',
-    (
-        r'^changelist\.xml$',
-        'changelist',
-        {
-            'sitemaps': sitemaps,
-            'section': 'changelist',
-            'template_name': 'mdstore/changelist.xml'
-        },
-    ),
-    (
-        r'^capabilitylist\.xml$',
-        'capabilitylist',
-    ),
-)
--- a/coda/coda_replication/urls.py
+++ b/coda/coda_replication/urls.py
@@ -4,9 +4,9 @@ from . import views
 urlpatterns = [
     url(r'^APP/queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue, name='app-detail'),
     url(r'^APP/queue/$', views.queue, name='app-list'),
-    url(r'^queue/$', views.queue_recent, name='index'),
-    url(r'^queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue_html, name='detail'),
-    url(r'^queue/search/$', views.queue_search, name='search'),
+    url(r'^queue/$', views.queue_recent, name='replication_index'),
+    url(r'^queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue_html, name='replication_detail'),
+    url(r'^queue/search/$', views.queue_search, name='replication_search'),
     url(r'^queue/search.json$', views.queue_search_JSON, name='search-json'),
-    url(r'^queue/stats/$', views.queue_stats, name='stats'),
+    url(r'^queue/stats/$', views.queue_stats, name='replication_stats'),
 ]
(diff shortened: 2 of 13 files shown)


coda/coda_mdstore/urls.py
```diff
@@ -31,42 +32,9 @@
     url(r'^about/$', views.about, name='about'),
     url(r'^robots.txt$', views.shooRobot, name='robots'),
     url(r'^feed/$', views.AtomSiteNewsFeed(), name='feed'),
+    url(r'^resourceindex\.xml$', sitemap_views.index),
+    url(r'^resourcelist-(?P<section>.+)\.xml$', sitemap_views.sitemap, name='sitemaps_views'),
+    url(r'changelist\.xml$', changelist),
```
ldko: resourceindex.xml, resourcelist-001.xml, and changelist.xml are all erroring for me with a `TypeError` about arguments. Maybe you need to pass `sitemaps` as an argument as it looks like was done before? If these URLs are also erroring for you and the tests didn't catch it, we should probably add tests that will show a failure before fixing.

coda/coda_replication/urls.py
```diff
@@ -4,9 +4,9 @@
 urlpatterns = [
     url(r'^APP/queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue, name='app-detail'),
     url(r'^APP/queue/$', views.queue, name='app-list'),
-    url(r'^queue/$', views.queue_recent, name='index'),
-    url(r'^queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue_html, name='detail'),
-    url(r'^queue/search/$', views.queue_search, name='search'),
+    url(r'^queue/$', views.queue_recent, name='replication_index'),
```
ldko: Since existing names use `-` not `_`, what do you think of changing all the `name` values you used `_` in to use `-` for consistency?

coda/templates/mdstore/bag_info.html
```diff
@@ -1,92 +1,92 @@
...
+
+{% endfor %}
+	</tbody>
+</table>
+
+
+{% else %}
+    <h3><i class="icon-remove icon-large"></i> There are no <em>premis events</em> associated with {{ bag }}.</h3>
+{% endif %}
+{% endblock %}
```
ldko: Do you know why this file shows so many changes even though it looks like there is only one line changed? Is it endline chars being switched or something?

madhulika95b: I made the autocrlf setting in my IDE before pushing this change. Not sure why it's showing so many changes. I'll look into it.

coda/coda_replication/urls.py
madhulika95b: sure! will do.

coda/coda_mdstore/urls.py
```diff
@@ -31,42 +32,9 @@
     url(r'^about/$', views.about, name='about'),
     url(r'^robots.txt$', views.shooRobot, name='robots'),
     url(r'^feed/$', views.AtomSiteNewsFeed(), name='feed'),
+    url(r'^resourceindex\.xml$', sitemap_views.index),
+    url(r'^resourcelist-(?P<section>.+)\.xml$', sitemap_views.sitemap, name='sitemaps_views'),
```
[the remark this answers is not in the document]
madhulika95b: I remember it was conflicting with something else. I will use a different name for it which will follow the trend.

coda/coda_mdstore/urls.py
madhulika95b: It seems like they are giving `TypeError` for me as well. And I didn't realize about it 'cos they weren't caught by the tests. I think adding new tests to check these is a good idea.

coda/coda_mdstore/urls.py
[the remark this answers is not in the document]
madhulika95b: Here, I'm importing `django.contrib.sitemaps.views` as `sitemap_views` and didn't want to name the url with the same name. I'm thinking to rename it to be `sitemap_views_sitemap`. Does it look ok or if y'all think it's weird I can change the alias name of import and use `sitemap_views` here.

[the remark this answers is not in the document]
ldko: To me `sitemap-views` suggests this is a reference to multiple views. I thought this represents just one view from the sitemaps views--but I could just be confused and getting it wrong. Did you confirm that using `name='sitemap'` is still causing a conflict?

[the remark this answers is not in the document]
ldko: or it could be `name=resourcelist`?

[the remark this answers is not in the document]
ldko: which issue? I just didn't like the name :D. It isn't going to fix the `TypeError`.

coda/coda_mdstore/urls.py
```diff
@@ -31,42 +32,29 @@
...
+        sitemap_views.index,
         {
             'sitemaps': sitemaps,
             'template_name': 'mdstore/resourceindex.xml'
         },
-    ),
-    (
-        r'^resourcelist-(?P<section>.+)\.xml$',
-        'sitemap',
+        ),
```
ldko: Very minor style thing: I would pull this `),` back one indent and do the same for the next couple `url` calls as well.

coda/coda_replication/urls.py
```diff
@@ -4,9 +4,9 @@
...
     url(r'^APP/queue/$', views.queue, name='app-list'),
-    url(r'^queue/$', views.queue_recent, name='index'),
-    url(r'^queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue_html, name='detail'),
-    url(r'^queue/search/$', views.queue_search, name='search'),
+    url(r'^queue/$', views.queue_recent, name='replication-index'),
+    url(r'^queue/(?P<identifier>ark:\/\d+\/.+?)/$', views.queue_html, name='replication-detail'),
+    url(r'^queue/search/$', views.queue_search, name='replication-search'),
     url(r'^queue/search.json$', views.queue_search_JSON, name='search-json'),
-    url(r'^queue/stats/$', views.queue_stats, name='stats'),
+    url(r'^queue/stats/$', views.queue_stats, name='replication_stats'),
```
ldko: I think we should change this to `replication-stats` for consistency.

coda/coda_validate/urls.py
```diff
@@ -5,10 +5,10 @@
 urlpatterns = [
     url(r'^APP/validate/(?P<identifier>.+?)/$', views.app_validate, name='app-detail'),
     url(r'^APP/validate/$', views.app_validate, name='app-list'),
-    url(r'^validate/$', views.index, name='index'),
+    url(r'^validate/$', views.index, name='validate_index'),
```
ldko: I think we should change this to `validate-index` for consistency.

coda/coda_validate/urls.py
```diff
@@ -5,10 +5,10 @@
...
     url(r'^validate/list/$', views.ValidateListView.as_view(), name='validate_list'),
     url(r'^validate/check.json', views.check_json, name='check_json'),
-    url(r'^validate/stats/$', views.stats, name='stats'),
+    url(r'^validate/stats/$', views.stats, name='validate_stats'),
```
ldko: II think we should change this to `validate-stats` for consistency.

coda/templates/mdstore/bag_info.html
```diff
@@ -1,92 +1,92 @@
...
```
ldko: I don't think we need that space on the end.

PR #131, opened 2019-09-13
Dj 11

madhulika95b:
@ldko @somexpert  Upgraded coda to django 1.11. Please review the changes.

### DIFF
--- a/coda/coda_replication/tests/test_views.py
+++ b/coda/coda_replication/tests/test_views.py
@@ -88,7 +88,7 @@ class TestQueueSearch:
         response = client.get(
             reverse('replication-search'))
 
-        page = response.context[-1]['entries']
+        page = response.context['entries']
         assert isinstance(page, Page)
 
     def test_blank_search_returns_all_queues(self, client):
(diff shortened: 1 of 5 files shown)


somexpert:
oh, right, with DEBUG set to True it should work fine with an empty ALLOWED_HOSTS in the local settings. But still, that's something we do want generally in the secrets template

madhulika95b:
I don't think ALLOWED_HOSTS is required in secrets.json file as it isn't required for the application to run but just for tests which have the setting `HTTP_HOST="example.com"`.

somexpert:
It may not be required when running things through docker, but if you were actually going to do a production deployment, the production settings do expect to retrieve the allowed hosts through the settings files. I _think_ we need to put it in the template, and then it will simply be left empty most of the time, but I could be misunderstanding things.

somexpert:
This is something we can deal with later in #132

ldko:
Do we need to run `makemigrations` since there was a minor change to models in one of the Django version upgrades?

madhulika95b:
Yes! I think I'll have to do that

ldko:
Ah, saw your Edit on the comment. No, we don't need to add that to the README. I realize now the likely reason it showed there were migrations to be created when I deployed this in Docker was because the requirements.txt was still using master of django-premis-event-service which didn't have the migration in it. When it uses the django 1.11 version, it will have it.

PR #135, opened 2019-10-25
Py2to3

madhulika95b:
@ldko @somexpert 
Upgraded coda to python3. Please review the changes.


coda/coda_mdstore/presentation.py
```diff
@@ -32,7 +31,7 @@ def getFileList(url):
     """
 
     fileList = []
-    handle = urllib2.urlopen(url)
+    handle = urllib.request.urlopen(url)
     soup = BSoup(handle)
     trList = soup.findAll('tr')
```
somexpert: So this has me a bit worried about our tests. The `findAll` method of BeautifulSoup 3 was changed in bs4 to be `find_all`. There are other changes as well that may apply to our code. We need to be careful to check for changes like that when upgrading dependencies beyond just a patch or minor version. For bs4, you can see a helpful guide [here](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#porting-code-to-bs4), but you may want to take this time to review the other dependency upgrades too to see if they require other modifications.

madhulika95b:
Ok. I will review the changes I've to make with regards to dependency upgrades. @ldko @somexpert Please hold on your review until I make sure of dependency upgrade changes.

coda/coda_mdstore/presentation.py
somexpert: I got curious as to whether it really _was_ working (contrary to the documentation) or if our tests weren't catching it. It seems we got the better result of it just being a failsafe in bs4 for just this kind of thing. So, the `findAll` still works, as it just points to the same `find_all` method as you can see [here](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py#L1303), but still, we don't have a guarantee on how long that will last. We should use the specified API for the version that we're using.

madhulika95b: I don't see tests testing that piece of code using `findAll` which is why tests didn't fail. I'll see and add tests for such pieces of code.

madhulika95b:
@ldko @somexpert 
I have rebased this branch on master where i have removed some unused views. This is ready for review.

coda/config/wsgi.py
```diff
@@ -14,15 +14,15 @@
...
+SITE_PACKAGES = os.path.join(ENV, '/lib/python3.7/site-packages')
 
 site.addsitedir(SITE_PACKAGES)
 sys.path.append(os.path.join(BASE_DIR, 'coda/'))
 
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'config.settings.production')
 
 activate_env = os.path.join(ENV, 'bin/activate_this.py')
-execfile(activate_env, dict(__file__=activate_env))
+exec(activate_env, dict(__file__=activate_env))
```
ldko: I am not sure that `exec` and `execfile` are an exact replacement. 
I think in other Django projects we are doing:
```
with open(activate_env) as activate_f:
    exec(compile(activate_f.read(), activate_env, 'exec'), dict(__file__=activate_env))
```

requirements.txt
```diff
@@ -1,20 +1,20 @@
...
+mysqlclient ==1.3.14
 python-dateutil==2.4
 requests==2.20.0
-pyoai==2.4.4
+pyoai==2.5.0
 pytest==3.10.1
 pytest-django==2.9.1
 flake8
 pytest-flake8
 mock
```
ldko: We should use the `unittest.mock` now.

coda/coda_mdstore/mdstore_oai.py
```diff
@@ -186,18 +186,19 @@ def makeDataRecord(
         raise Exception("dcDict is empty")
     setList = []
     header = common.Header(
+        bagObject,
```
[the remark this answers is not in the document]
madhulika95b: I think I chose to use bagObject here, cos we are loading those functions (common.Header, common.Metadata) with `bagObject`'s attributes like Id, date and the element in `pyoai/common.py` would refer to bagObject.

coda/coda_mdstore/presentation.py
```diff
@@ -32,20 +26,20 @@ def getFileList(url):
...
+            anchors = td.find_all('a')
             for anchor in anchors:
                 try:
                     # if anchor.contents and "Parent Directory" in contents:
                     if anchor['href'][-1] == "/":
                         continue
                     fileList.append(anchor['href'])
-                except Exception, e:
+                except Exception as e:
                     raise e
```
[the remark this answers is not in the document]
madhulika95b: It's the way the code has been written in python2. I think I haven't noticed it before. I can catch a specific Exception here.

coda/config/wsgi.py
madhulika95b: Well...I wasn't sure about this change myself. But since the setting worked fine and the app was working I thought I'll let it be there. But I can change it to what you have suggested if we are following that in other django projects.

coda/coda_validate/models.py
```diff
@@ -60,11 +60,14 @@ class Meta:
...
-        sql = """select date(last_verified) as dt, count(date(last_verified)) as ct
-        from %s
-        where last_verified_status != 'Unverified'
-        group by date(last_verified)
-        order by last_verified asc""" % (cls._meta.db_table,)
+        sql = """select dt, count(dt)
+                from
+                    (select date(last_verified) as dt
+                    from %s
+                    where last_verified_status != 'Unverified') as res
```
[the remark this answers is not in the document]
madhulika95b: Not anymore! I'll remove it.

coda/coda_mdstore/resourcesync.py
```diff
@@ -45,8 +46,8 @@ def index(
     req_site = get_current_site(request)
 
     sites = []
-    for section, site in sitemaps.iteritems():
-        if callable(site):
+    for section, site in sitemaps.items():
+        if isinstance(site, collections.Callable):
```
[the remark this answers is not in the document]
madhulika95b: Oh...ok. I'll revert theses changes.

coda/coda_mdstore/presentation.py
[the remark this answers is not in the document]
ldko: My question isn't about a specific exception being raised, it is about the act of catching an exception only to then raise it as is. It seems like that is the same thing that would happen if you don't catch the exception at all; the exception would raise the same.

coda/config/wsgi.py
ldko: I don't think this file gets used in dev/Docker environment, in which case you wouldn't see it not working. If I open a Python 3 interpreter and run `exec(activate_env, dict(__file__=activate_env))` where `activate_env` is a string path as it is above, I get an exception. [The docs](https://docs.python.org/3.3/whatsnew/3.0.html?highlight=execfile#builtins) indicate the first argument should be a file object in this case.

coda/coda_validate/models.py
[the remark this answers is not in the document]
ldko: ok, thanks for checking.

coda/coda_mdstore/mdstore_oai.py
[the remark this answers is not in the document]
ldko: Since bagObject is not an `Element` type, I think we should pass `None` instead if it works the same.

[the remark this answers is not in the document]
madhulika95b: I'll try it and see.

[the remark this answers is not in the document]
ldko: Ok. I did just check aubrey since I couldn't find pyoai documentation to inform on it other than the code, and that is how we changed it there (`None` as the first argument) for the upgrade.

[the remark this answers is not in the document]
madhulika95b: I tried using `None` and the app seems to work fine. Thanks for checking with the aubrey upgrade.

coda/coda_mdstore/presentation.py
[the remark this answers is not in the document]
madhulika95b: Maybe the intention here was to write except clause just for syntax purpose and raised the exception which will happen anyway, instead of doing nothing .
Do you think we can change it to 
```
                try:
                    # if anchor.contents and "Parent Directory" in contents:
                    if anchor['href'][-1] == "/":
                        continue
                    fileList.append(anchor['href'])
                except specific_exception:
                    pass
```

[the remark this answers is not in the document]
ldko: Theoretically this could throw a `KeyError` or an `IndexError`, but based on the context of the pages we are using this for, if those occur, there is something unexpectedly wrong with the page, so it might make sense to keep raising an exception. If you aren't comfortable with removing the try/except lines, you can leave it as is.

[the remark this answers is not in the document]
madhulika95b: I'd prefer removing try/except clauses from here.

coda/coda_mdstore/views.py
```diff
@@ -1,9 +1,8 @@
-import codecs
 import copy
 import uuid
 
-from urllib2 import urlopen
-from urllib import urlencode
+from urllib.request import urlopen
+from urllib.parse import urlencode
 try:
```
somexpert: We should just automatically start replacing these try/except blocks with `import json`.

coda/coda_replication/views.py
```diff
@@ -434,6 +432,6 @@ def queue_list(request):
         return HttpResponseBadRequest('Page does not exist.')
 
     feedText = etree.tostring(atomFeed, pretty_print=True)
-    feedText = '<?xml version="1.0"?>\n{0}'.format(feedText)
+    feedText = b'<?xml version="1.0"?>\n %s' % feedText
```
somexpert: We should replace `%s` with `%b` to avoid any confusion.

somexpert: In fact, that should be done in a few other places here too. Currently `%s` _is_ an alias for `%b`, but we should be using `%b` for clarity, as we're making a bytes object, not a str.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,60 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-    def test_response_content(self):
-        assert 0
+        # Mock file names found at the bag's root level.
+        with mock.patch('coda_mdstore.views.getFileList',
+                        return_value=['bagit.txt', 'bag-info.txt']):
+            response = views.bagURLList(request, bag.name)
+
+        assert (b'https://coda/bag-info.txt\n'
+                b'https://coda/bagit.txt\n'
+                b'https://coda/data/file02.txt\n'
```
somexpert: This is really tiny, but I don't think we really need to check for the two `data/` files here, since that seems to be the point of the previous test.

madhulika95b: Well..I don't HAVE to test both the `data/` files, bit I did, 'cos I'm making them both as side_effect for mocked `getFileHandle` and asserting both will make it easier to read the test.

coda/coda_replication/views.py
madhulika95b: I will make these changes.

coda/coda_mdstore/views.py
madhulika95b: Will get rid of it.

coda/coda_mdstore/tests/test_views.py
somexpert: fair enough :thumbsup:

requirements.txt
```diff
@@ -1,20 +1,19 @@
...
-pyoai==2.4.4
+pyoai==2.5.0
 pytest==3.10.1
 pytest-django==2.9.1
 flake8
 pytest-flake8
-mock
 factory_boy
-codalib==1.0.3
 sqlparse==0.2.4
```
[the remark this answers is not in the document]
madhulika95b: I have actually only upgraded those packages which aren't compatible anymore and left the others as is. So, I can try upgrading them and see if they need any code changes and go back if they do and make an issue on those upgrades. Does it sound ok!?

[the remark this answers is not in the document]
somexpert: I think that sounds reasonable

coda/coda_replication/models.py
```diff
@@ -1,7 +1,7 @@
 from django.db import models
 # this is quite silly:
 # from django.conf import settings
-# import simplejson
+# import json
 # try:
 #     status_choice_json = open(settings.STATUS_CHOICE_FILE_PATH).read()
 #     status_choice_dict = simplejson.loads(status_choice_json)
```
somexpert: Need to replace simplejson with json here too

madhulika95b: yea...just in case if that code is brought back again..it will not have the unnecessary try..except

somexpert: I don't think that try/except block is like the others, I _think_ that's catching something else, but yeah, we do need to just do the simple replacement

madhulika95b: oh...yea...there's no try except here for the import. It'll error if it comes back.

ldko: Is there a reason we don't just delete this block of commented out code?

somexpert: heh, I guess that should have been the first question here :smile:

madhulika95b: I think it can be deleted. (But didn't delete it just in case if it's supposed to be brought back later.)

ldko: Let's delete it. I think it has been sitting there commented out for years.

PR #136, opened 2019-10-30
Add more tests

madhulika95b:
@ldko @somexpert 
Updated findAll() used in beautifulsoup and added some tests which test the piece of code using findAll(). Also added a couple tests by mocking `lxml.etree.XML()` to make sure that exception is caught properly in `coda_validate.views`. Please review them.

### DIFF
--- a/coda/coda_mdstore/tests/test_utils.py
+++ b/coda/coda_mdstore/tests/test_utils.py
@@ -603,3 +603,22 @@ class TestUpdateBag:
         update_bag_infos = updated_bag.bag_info_set.all()
         assert old_bag_info1.field_name not in [b.field_name for b in update_bag_infos]
         assert old_bag_info2.field_name not in [b.field_name for b in update_bag_infos]
+
+
+@mock.patch('coda_mdstore.presentation.urllib.request.urlopen')
+def test_getFileList(mock_urlopen):
+    """ Test all href tags are extracted as files."""
+    text = """<html>
+                 <body>
+                 <tr> <td>test</td> <td>data</td> </tr>
+                 <tr> <td>of </td> </tr>
+                 <tr> <td>
+                    <a href='bag-info.txt'>url</a>
+                    <a href='manifest-md5.txt'>here</a>
+                    <a href='bagit.txt'>here</a>
+                 </td> </tr>
+                 </body>
+              </html>"""
+    mock_urlopen.return_value = text
+    filelist = presentation.getFileList('https://coda/testurl')
+    assert ['bag-info.txt', 'manifest-md5.txt', 'bagit.txt'] == filelist
--- a/coda/coda_mdstore/tests/test_utils.py
+++ b/coda/coda_mdstore/tests/test_utils.py
@@ -607,8 +607,8 @@ class TestUpdateBag:
 
 @mock.patch('coda_mdstore.presentation.urllib.request.urlopen')
 def test_getFileList(mock_urlopen):
-    """ Test all href tags are extracted as files."""
-    text = """<html>
+    """Test all attribute values are extracted as files."""
+    text = b"""<html>
                  <body>
                  <tr> <td>test</td> <td>data</td> </tr>
                  <tr> <td>of </td> </tr>
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -592,17 +592,21 @@ class TestBagURLListView:
             views.bagURLList(request, bag.name)
 
     def test_output_text(self, rf):
+        """Test the outputText obtained from mocked getFileHandle object"""
         self.getFileHandle.return_value.url = 'https://coda/testurl'
-        self.getFileHandle.return_value.readline.side_effect = ['This is sample testing text', '']
+        self.getFileHandle.return_value.readline.side_effect = [
+            '<html, class=coda lang=en dir=ltr>', '']
         bag = FullBagFactory.create()
         request = rf.get('/')
         response = views.bagURLList(request, bag.name)
-        assert b'https://coda/is sample testing text' in response.content
+        assert b'https://coda/class=coda lang=en dir=ltr>' in response.content
         assert response.status_code == 200
 
     def test_top_files_block(self, rf):
+        """Test getFileHandle and getFileList methods are called without any failures"""
         self.getFileHandle.return_value.url = 'https://testurl'
-        self.getFileHandle.return_value.readline.side_effect = ['Sample text', '']
+        self.getFileHandle.return_value.readline.side_effect = [
+            '<html, class=coda lang=en dir=ltr>', '']
         bag = FullBagFactory.create()
         request = rf.get('/')
         text = """<html>
@@ -615,9 +619,11 @@ class TestBagURLListView:
                      </td> </tr>
                      </body>
                   </html>"""
+        # Mock return value of getFileHandle method
         with mock.patch('coda_mdstore.presentation.urllib.request.urlopen', return_value=text):
             response = views.bagURLList(request, bag.name)
-        assert b'https://testurl.com\nhttps://test.com\nhttps://text' in response.content
+        assert b'https://testurl.com\nhttps://test.com\nhttps://class=coda lang=en dir=ltr>' in \
+               response.content
         assert response.status_code == 200
 
     @pytest.mark.xfail(reason='Refactor is required.')
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -591,31 +591,38 @@ class TestBagURLListView:
         with pytest.raises(http.Http404):
             views.bagURLList(request, bag.name)
 
-    # @pytest.mark.xfail(reason='Refactor is required.')
     @mock.patch('coda_mdstore.views.len', return_value=5)
     def test_output_text(self, rf):
         self.getFileHandle.return_value.url = 'https://coda/testurl'
-        self.getFileHandle.return_value.readline.return_value = 'This is sample testing text'
+        self.getFileHandle.return_value.readline.side_effect = ['This is sample testing text', '']
+        bag = FullBagFactory.create()
+        request = rf.get('/')
+        response = views.bagURLList(request, bag.name)
+        assert b'https://coda/is sample testing text' in response.content
+        assert response.status_code == 200
+
+    def test_top_files_block(self, rf):
+        self.getFileHandle.return_value.url = 'https://testurl'
+        self.getFileHandle.return_value.readline.side_effect = ['Sample text', '']
         bag = FullBagFactory.create()
         request = rf.get('/')
         text = """<html>
-                    <body>
-                    <tr> <td>test</td> <td>data</td> </tr>
-                    <tr> <td>of </td> </tr>
-                    <tr> <td> <a href='test.com'>url</a> <a href='testurl.com'>here</a> </td> </tr>
-                    </body>
-                </html>"""
+                     <body>
+                     <tr> <td>test</td> <td>data</td> </tr>
+                     <tr> <td>of </td> </tr>
+                     <tr> <td>
+                        <a href='test.com'>url</a>
+                        <a href='testurl.com'>here</a>
+                     </td> </tr>
+                     </body>
+                  </html>"""
         with mock.patch('coda_mdstore.presentation.urllib.request.urlopen', return_value=text):
             response = views.bagURLList(request, bag.name)
+        assert b'https://testurl.com\nhttps://test.com\nhttps://text' in response.content
         assert response.status_code == 200
 
     @pytest.mark.xfail(reason='Refactor is required.')
-    def test_top_files_block(self):
-        # TODO: rename once implemented.
-        assert 0
-
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_path_is_not_unicode_safe(self):
+    def test_path_is_not_unicode_safe(self, rf):
         assert 0
 
     @pytest.mark.xfail(reason='Refactor is required.')
@@ -655,10 +662,26 @@ class TestBagURLListScrapeView:
         with pytest.raises(http.Http404):
             views.bagURLListScrape(request, 'ark:/00001/id')
 
-    @pytest.mark.xfail(reason='pairtreeCandidateList is not available '
-                              'in scope.')
-    def test_response_content(self):
-        assert 0
+    def test_response_content(self, rf):
+        bag = FullBagFactory.create()
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        request = rf.get('/')
+        text = """<html>
+                    <body>
+                    <tr> <td>test</td> <td>data</td> </tr>
+                    <tr> <td>of </td> </tr>
+                    <tr> <td>
+                        <a href='test.com'>url</a>
+                        <a href='testurl.com'>here</a>
+                    </td> </tr>
+                    </body>
+                  </html>"""
+
+        with mock.patch('coda_mdstore.presentation.urllib.request.urlopen', return_value=text):
+            response = views.bagURLListScrape(request, bag.name)
+
+        assert b'test.com\ntesturl.com' == response.content
+        assert response.status_code == 200
 
 
 class TestBagFullTextSearchHTMLView:
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -612,8 +612,8 @@ class TestBagURLListView:
         self.getFileHandle.return_value.url = 'https://coda/testurl'
         # Mock what gets read from the manifest file.
         self.getFileHandle.return_value.readline.side_effect = [
-            b'<html coda/bag/123/manifest-md5.txt',
-            b'<html coda/bag/123/bagit.txt',
+            b'192e635b17a9c2aea6181f0f87cab05d  data/file01.txt',
+            b'18b7c500ef8bacf7b2151f83d28e7ca1  data/file02.txt',
             b'']
         bag = FullBagFactory.create()
         request = rf.get('/')
@@ -624,8 +624,8 @@ class TestBagURLListView:
 
         assert (b'https://coda/bag-info.txt\n'
                 b'https://coda/bagit.txt\n'
-                b'https://coda/coda/bag/123/bagit.txt\n'
-                b'https://coda/coda/bag/123/manifest-md5.txt') in response.content
+                b'https://coda/data/file02.txt\n'
+                b'https://coda/data/file01.txt') in response.content
         assert response.status_code == 200
 
 
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -594,6 +594,7 @@ class TestBagURLListView:
     def test_output_text(self, rf):
         """Test the outputText obtained from mocked getFileHandle object"""
         self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock the return value of urllib.request.urlopen(url) in getFileHandle()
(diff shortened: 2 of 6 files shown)


coda/coda_validate/tests/test_utils.py
```diff
@@ -64,9 +65,11 @@ def test_xmlToValidateObject(validate_feed):
     assert validate_xml.server == validate.server
 
 
-@pytest.mark.xfail(reason='Exception will never be raised directly from the function.')
-def test_xmlToValidateObject_raises_exception():
-    assert 0
+@mock.patch('coda_validate.views.etree.XML', return_value=None)
+def test_xmlToValidateObject_raises_exception(mock_xml):
+    with pytest.raises(Exception) as e:
```
somexpert: We should probably catch a specific exception here and in the other similar test.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,14 +591,34 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_output_text(self):
-        assert 0
+    def test_output_text(self, rf):
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        self.getFileHandle.return_value.readline.side_effect = ['This is sample testing text', '']
+        bag = FullBagFactory.create()
+        request = rf.get('/')
+        response = views.bagURLList(request, bag.name)
+        assert b'https://coda/is sample testing text' in response.content
```
somexpert: Maybe we should change the string returned by readline to be something that more closely resembles the kind of string that would actually be returned, because on first glance, I don't understand why it is desirable for the first word to have been removed.

somexpert: I think this sort of applies to the next two added tests too. It might make these tests easier to read through if the returned and mocked values are constructed to make it clearer what the expected result should be.

ldko: Comments can also help with making tests easier to read through and understand.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -641,10 +661,26 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
+                    <tr> <td>of </td> </tr>
+                    <tr> <td>
+                        <a href='test.com'>url</a>
+                        <a href='testurl.com'>here</a>
+                    </td> </tr>
+                    </body>
+                  </html>"""
+
+        with mock.patch('coda_mdstore.presentation.urllib.request.urlopen', return_value=text):
+            response = views.bagURLListScrape(request, bag.name)
```
ldko: Regarding testing `bagURLListScrape` it looks to me like we aren't actually using it and it and  `pairtreeCandidateList` could be removed from the code. I was comparing this test to `bagURLListScrape` and noticed that `bagURLListScrape` calls `getFileHandle(identifier, "", pairtreeCandidateList)`. This should error if it were being used because `getFileHandle` only takes 2 arguments, not 3. I see that `getFileHandle` was previously mocked and unfortunately the way it was done hides that this function is broken. If you agree it can be safely removed, we should make it go away :).

madhulika95b: Yea...I don't see those views are being used and I think they can be removed.

madhulika95b:
I'm not sure what causes `UnicodeDecodeError` [here](https://github.com/unt-libraries/coda/blob/add_more_tests/coda/coda_mdstore/views.py#L588). I can add a test for the `except UnicodeDecodeError` piece of code there.

ldko:
https://github.com/unt-libraries/coda/commit/eda91bfc9a66890d935687b5e01a6f54d50f8a22#diff-ebed849629179ae4a0abd56f966a1cacL588 it used to call `unicode` function which you removed in a py2to3 commit. We don't need the try/except or assigning to `unipath` anymore.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
+                     <body>
+                     <tr> <td>test</td> <td>data</td> </tr>
+                     <tr> <td>of </td> </tr>
+                     <tr> <td>
+                        <a href='test.com'>url</a>
+                        <a href='testurl.com'>here</a>
+                     </td> </tr>
+                     </body>
+                  </html>"""
+        # Mock urlopen() in getFileHandle()
```
ldko: I don't think this comment is needed as it doesn't add any clarification to the code--it is a pretty direct translation.

madhulika95b: How about  `Mock return value of getFileHandle()`? or do you think its not just necessary here?

coda/coda_validate/tests/test_utils.py
somexpert: So I didn't realize at the time that you _were_ catching as specific an exception as you could. Odd that we'd raise a plain Exception. So, while my intention wasn't to change the called code, maybe this is a good thing? I can't image there being any unforeseen consequences of raising (and catching) a more specific exception, as anything that was looking for the base Exception class should still catch this more specific one, but still, thoughts @ldko, @madhulika95b? Is this a safe change?

madhulika95b: well.. I tried raising `ParseError`, `ValueError` exceptions in the source code and it didn't raise any exception but `ValidationError` raised the exception. And mostly these exceptions are raised when dealing with xml data parsing. So I thought this would be safe. Do you think I should add any more exceptions here?

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_top_files_block(self):
-        # TODO: rename once implemented.
-        assert 0
-
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_path_is_not_unicode_safe(self):
-        assert 0
+    def test_output_text(self, rf):
+        """Test the outputText obtained from mocked getFileHandle object"""
```
somexpert: I think this docstring doesn't really clarify anything. There's no need to test the output from a mocked object, as we control exactly what a mocked object returns.

coda/coda_validate/tests/test_utils.py
somexpert: The specific exception raised doesn't bother me, I just hadn't made this comment with the expectation that it would affect the code being tested. I originally thought there _already_ was a more specific subclass of Exception being raised, and that we should catch whatever that was. But I see that's not the case. Had I known that at the time, I wouldn't have commented at all, but since the change has been made, I'm just wanting to make sure that it's a _safe_ change, which I believe it is, but I wanted to get other opinions. I tend to be very cautious about changing the code under test rather than _just_ changing the test itself.

coda/coda_mdstore/tests/test_views.py
madhulika95b: I think I can rephrase it as `Test the outputText obtained contains mocked getFileHandle return_value`, 'cos this test checks if the mocked object's return_value gets processed correctly and goes to `outputText` variable. What do you think?

coda/coda_validate/tests/test_utils.py
madhulika95b: Yes, what you said is totally true. 'cos catching specific exceptions is a good habit generally, I thought it's worth changing it. Let's see what @ldko has to say abt it.

coda/coda_mdstore/tests/test_views.py
somexpert: I think the strings themselves are better now, but yes, some descriptive comments would go a long ways towards making this more understandable too. I think the focus of these comments should not be on describing the lines of code themselves, but rather on why things are happening. What is special about the input in one test vs another? What makes one different? Why is the resulting value correct? Comments that answer those questions can be really helpful for tests that aren't showing something obvious.

madhulika95b: After discussing about what the tests are doing with @somexpert , I'm renaming the test names and modifying the doc strings for readability.

coda/coda_validate/tests/test_utils.py
ldko: > well.. I tried raising ParseError, ValueError exceptions in the source code and it didn't raise any exception but ValidationError raised the exception. And mostly these exceptions are raised when dealing with xml data parsing. So I thought this would be safe. Do you think I should add any more exceptions here?

The code that has been changed in `codavalidate/views.py` is to explicitly raise a certain type of exception, so I am not sure why that wouldn't raise an exception. I think it is ok to raise a more specific exception, but I don't think [`ValidationError`](https://docs.djangoproject.com/en/1.11/ref/exceptions/#validationerror), a Django exception  "raised when data fails form or model field validation" is necessarily the right choice. I think `ValueError` makes a little more sense, which is why I am wondering about the comment saying it didn't raise any exception.

madhulika95b: Umm...I tried changing the exception now, and it worked. But I got error when I tried using ValueError before. I guess my test used cached test code which didn't match with the raised exception. But I can use `ValueError` in raising exception.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_top_files_block(self):
-        # TODO: rename once implemented.
-        assert 0
-
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_path_is_not_unicode_safe(self):
-        assert 0
+    def test_response_content(self, rf):
+        """Test the output contains domain and parsed tag attributes from getFileHandle."""
```
ldko: I am confused by this docstring. I am not sure from where parsed tag attributes is coming into this. This test shows itself to be about testing the response from the `bagURLList` view.

madhulika95b: `bagURLList` uses `getFileList` which parses each `href` tag as a file and adds it to pathList.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
```
ldko: Based on the arguments used in calling `getFileHandle` from `bagURLList` (` handle = getFileHandle(identifier, "manifest-md5.txt")`), I think this would make more sense with a value ending in `manifest-md5.txt` as should be the case in practice. Something like `http://coda/bag/123/manifest-md5.txt`.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
+        # Mock the return value of urllib.request.urlopen(url) in getFileHandle()
```
ldko: Maybe `# Mock what gets read from the manifest file.`

coda/coda_mdstore/tests/test_views.py
ldko: Maybe for a clearer idea of what is happening in the function under test it should be a side effect and return `['http://coda/bag/123/manifest-md5.txt', 'http://coda/bag/123/']`

coda/coda_mdstore/tests/test_views.py
ldko: I believe this test can be deleted.

madhulika95b: I deleted this test and its corresponding view and merged it to master. Once this branch gets merged to py2to3 branch I'll do a rebase.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,35 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-        assert 0
+    def test_response_content_has_topFiles(self, rf):
+        """Test topFiles are returned in the response."""
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock what gets read from the manifest file.
+        self.getFileHandle.return_value.readline.side_effect = [
+            '<html coda/bag/123/manifest-md5.txt', '<html coda/bag/123/bagit.txt', '']
+        bag = FullBagFactory.create()
+        request = rf.get('/')
+        with mock.patch('coda_mdstore.views.getFileList',
```
ldko: You could add a comment like "Mock file names found at the bag's root level."

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,35 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
+                        return_value=['bagit.txt', 'bag-info.txt']):
+            response = views.bagURLList(request, bag.name)
 
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_returns_status_code_200(self):
-        assert 0
+        assert b'https://coda/bag-info.txt\nhttps://coda/bagit.txt\nhttps://coda/coda/' \
+               b'bag/123/bagit.txt\nhttps://coda/coda/bag/123/manifest-md5.txt' in response.content
```
ldko: For easier understanding of the test, I would format this like:
```
assert (b'https://coda/bag-info.txt\n'
        b'https://coda/bagit.txt\n'
        b'https://coda/coda/bag/123/bagit.txt\n'
        b'https://coda/coda/bag/123/manifest-md5.txt') in response.content
```

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,35 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
 
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_path_is_not_unicode_safe(self):
-        assert 0
+    def test_response_content_has_topFiles(self, rf):
+        """Test topFiles are returned in the response."""
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock what gets read from the manifest file.
+        self.getFileHandle.return_value.readline.side_effect = [
+            '<html coda/bag/123/manifest-md5.txt', '<html coda/bag/123/bagit.txt', '']
```
ldko: Again, for easier reading of the test,  I would change the `side_effect` values to be like mentioned above.

coda/coda_mdstore/tests/test_utils.py
```diff
@@ -603,3 +603,22 @@ def test_existing_bag_info_objects_are_update(self, bag_xml, rf):
         update_bag_infos = updated_bag.bag_info_set.all()
         assert old_bag_info1.field_name not in [b.field_name for b in update_bag_infos]
         assert old_bag_info2.field_name not in [b.field_name for b in update_bag_infos]
+
+
+@mock.patch('coda_mdstore.presentation.urllib.request.urlopen')
+def test_getFileList(mock_urlopen):
+    """ Test all href tags are extracted as files."""
```
ldko:
```suggestion
    """Test all href attribute values are extracted as files."""
```

coda/coda_mdstore/views.py
```diff
@@ -583,21 +583,13 @@ def bagURLList(request, identifier):
...
     for path in pathList:
-        # path = urllib2.quote(path)
-        try:
-            unipath = path
-        except UnicodeDecodeError:
-            try:
-                unipath = utf8_decoder(path)[0]
-            except UnicodeDecodeError:
-                unipath = latin_decoder(path)[0]
         # CODA_PROXY_MODE is a settings variable
```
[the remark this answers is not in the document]
madhulika95b: The `fileHandle` returned by `urlopen` returns its content in `bytes`. And `bs4.element.ResultSet` returned by soup.find_all method is `str`, so `fileList` returned contains its contents as `str`.  So, `pathList` is taking both `str` coming from fileList and `bytes` objects coming from fileHandle. 

I think before appending fileHandle contents to `pathList` they can be decoded and we can then add the check to see if `pathList` has any elements in `bytes` as you suggested.

```
    while line:
        line = line.strip()
        parts = line.split(None, 1)
        if len(parts) > 1:
            pathList.append(parts[1].decode())
        line = handle.readline()
```

coda/coda_mdstore/views.py
```diff
@@ -583,21 +583,19 @@ def bagURLList(request, identifier):
...
+        if isinstance(path, bytes):
+            try:
+                path = path.decode()
+            except:
```
ldko: We should catch a specific exception here.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,35 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
 
-    @pytest.mark.xfail(reason='Refactor is required.')
-    def test_output_text(self):
-        assert 0
+    def test_response_content(self, rf):
+        """Test the output contains domain and parsed tag attributes from getFileHandle."""
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock what gets read from the manifest file.
+        self.getFileHandle.return_value.readline.side_effect = [
+            '<html coda/bag/123/manifest-md5.txt', '<html coda/bag/123/bagit.txt', '']
```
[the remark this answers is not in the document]
ldko: Any further thoughts on making the returned text look more like the text file as shown in my previous comment?

coda/coda_mdstore/views.py
ldko: Also, we can now remove:
```
import codecs
```
and
```
utf8_decoder = codecs.getdecoder("utf-8")
latin_decoder = codecs.getdecoder("latin_1")
```

coda/coda_mdstore/tests/test_views.py
ldko: this one too?

coda/coda_mdstore/tests/test_views.py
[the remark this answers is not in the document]
madhulika95b: I think I got confused, instead of using the urls in "manifest-md5.txt", I used the url which kind of looks like the url for 'manifest-md5.txt'.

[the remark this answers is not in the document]
ldko: Also, I really would like to get rid of the `<html` bit, because it is a text file, so that is why I put in checksums like what would be in the manifest text file.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -591,22 +591,42 @@ def test_raises_http404_file_handle_is_falsy(self, rf):
...
-    def test_returns_status_code_200(self):
-        assert 0
+    def test_response_content_has_topFiles(self, rf):
+        """Test topFiles are returned in the response."""
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock what gets read from the manifest file.
+        self.getFileHandle.return_value.readline.side_effect = [
+            b'<html coda/bag/123/manifest-md5.txt',
+            b'<html coda/bag/123/bagit.txt',
+            b'']
```
ldko: This one could be changed to match the previous test.

madhulika95b: well...I didn't change it 'cos this test is written to test top files part of the code where we have `topFileHandle = getFileHandle(identifier, "")`. So, here we aren't opening the maifest.txt file.

ldko: If that were the case, than this would not be in the response, right?
```
                b'https://coda/coda/bag/123/bagit.txt\n'
                b'https://coda/coda/bag/123/manifest-md5.txt') in response.content
```

madhulika95b: They are coming from mocked `getFileList`. I think I can change `bag-info.txt` to be `manifest-md5.txt`.

PR #143, opened 2020-05-20
Fix ambiguous var

ldko:
This is to fix a Travis build error for ambiguous variable named 'l'. It also makes the recommended version bump of httplib2.


ldko:
This is ready for review @madhulika95b @somexpert . It is to fix a couple very small things for merging into py3  before merging into master.

somexpert:
I appreciate that flake8 added this. LGTM :thumbsup:

PR #148, opened 2020-09-02
Django2.2 upgrade

### LINKED ISSUE #147: Upgrade to Django 2.2 LTS

madhulika95b:
@ldko 
This upgrades coda to Django 2.2 and closes #147


coda/coda_mdstore/presentation.py
```diff
@@ -86,8 +85,7 @@ def bagSearch(bagString):
     return a list of related bags
     """
 
-    bagList = Bag.objects.filter(
-        bag_info__field_body__search=bagString).distinct()
+    bagList = Bag.objects.filter(bag_info__field_body__exact=bagString).distinct()
```
ldko, line 88: I know the built-in `__search` lookup has been removed, and I see that there is an error if you try to use the replacement lookup you defined in this PR, but I don't think using the `__exact` lookup here is going to work for our use case. I believe it changes the behavior of the searching to require the user to, for instance, need to know the full description content of a bag and search for that in its entirety rather than a few keywords.

coda/config/settings/base.py
```diff
@@ -74,7 +74,8 @@ def get_secret(setting, secrets=secrets):
         'HOST': get_secret("DB_HOST"),
         'PORT': get_secret("DB_PORT"),
         'OPTIONS': {
-            'init_command': 'SET default_storage_engine=MyISAM'
+            'init_command': 'SET default_storage_engine=MyISAM',
+            'sql_mode': 'traditional',
```
ldko, line 78: Noting that we discussed not using 'traditional', since we are using a non-transactional storage engine.

coda/coda_mdstore/presentation.py
madhulika95b: Oh..I intended to use `__search` here. It is here by mistake. Can you tell me more about the error with the replacement lookup?

ldko: Yeah, actually there isn't an error! I was messing with the code locally the other day and forgot and had caused the error. I think we just need to switch to use __search here like you intended.

ldko (approved):
works for me :+1:

PR #149, opened 2020-09-02
Use django-pes 2.0.0 version.

madhulika95b:
This PR uses django 1.11 dependency of django-pes.


ldko (approved):
Thank you, Madhulika!

Issue #145, opened 2020-08-13
Adjust default search behavior of externalIdentifierSearchJSON


vphill:
The default behavior of the `extidentifier.json` api is to return entries for the value supplied in the `ark` parameter.  Currently, it returns all instances of that identifier, so if there are multiple versions in the database from changes being made to the object over time they are all returned. 

`extidentifier.json?ark=metapth1234567`

This view is here in the code
https://github.com/unt-libraries/coda/blob/d8c34b641b4429bab8b502466ccbe767c37bbd5b/coda/coda_mdstore/views.py#L670

I would like to adjust this view to only show the most recently added item, based on bagging_date instead of default showing all of the items. 

We can add a second parameter `showAll` that has a default value of `False` and if `True` then it would replicate the current behavior of showing all versions of that identifier in the coda repository.

PR #150, opened 2020-09-09
Remove duplicates from extidenifier.json results

### LINKED ISSUE #146: extidenifier.json returning duplicate entries
I'm not sure that the `.distinct()` ORM call in the `externalIdentifiersSearchJSON` view is actually working correctly. 

See examples like these. 

/extidentifier.json?ark=metapth123456
/extidentifier.json?ark=metapth6726

I think the duplicate entries in the External_Identifiers table are a holdover and may get deleted if a full reindex of coda is performed.   Even with that, we should look at why duplicate entries are returned.

madhulika95b:
@ldko 

This closes #146 #145 
`distinct()` on filtered `ExternalIdentifierFactory` objects doesn't seem to remove duplicates from `externalIdentifierSearchJSON` view. Probably `distinct()` is considering them as distinct objects because they are all given unique ids. So, I added the same logic used in `externalIdentifierSearch` view to remove duplicates.

### DIFF
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -675,16 +675,19 @@ def externalIdentifierSearchJSON(request):
 
     identifiers = (External_Identifier.objects
                                       .select_related('belong_to_bag')
-                                      .filter(value=ark)
-                                      .distinct())
-    data = [
-        {
-            'name': exid.belong_to_bag.name,
-            'oxum': exid.belong_to_bag.oxum,
-            'bagging_date': exid.belong_to_bag.bagging_date.strftime('%Y-%m-%d')
-        }
-        for exid in identifiers
-    ]
+                                      .filter(value=ark))
+    unique_identifiers = []
+    data = []
+    for bagInfoObject in identifiers:
+        bag = bagInfoObject.belong_to_bag
+        if bag not in unique_identifiers:
+            unique_identifiers.append(bag)
+            data.append(
+                {
+                    'name': bag.name,
+                    'oxum': bag.oxum,
+                    'bagging_date': bag.bagging_date.strftime('%Y-%m-%d')
+                })
 
     data = json.dumps(data, indent=4, sort_keys=True)
     return HttpResponse(data, content_type='application/json')
--- a/coda/coda_mdstore/views.py
+++ b/coda/coda_mdstore/views.py
@@ -670,24 +670,44 @@ def externalIdentifierSearch(request, identifier=None):
 def externalIdentifierSearchJSON(request):
     """External_Identifier search formatted in JSON."""
     ark = request.GET.get('ark', '')
+    showAll = request.GET.get('showAll', False)
     if ('ark:/%d' % settings.ARK_NAAN) not in ark:
         ark = 'ark:/%d/%s' % (settings.ARK_NAAN, ark)
-
-    identifiers = (External_Identifier.objects
-                                      .select_related('belong_to_bag')
-                                      .filter(value=ark))
-    unique_identifiers = []
     data = []
-    for bagInfoObject in identifiers:
-        bag = bagInfoObject.belong_to_bag
-        if bag not in unique_identifiers:
-            unique_identifiers.append(bag)
-            data.append(
+    if not showAll:
+        identifiers = (External_Identifier.objects
+                                          .select_related('belong_to_bag')
+                                          .filter(value=ark)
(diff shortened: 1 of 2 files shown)


coda/coda_mdstore/views.py
```diff
@@ -675,16 +675,19 @@ def externalIdentifierSearchJSON(request):
...
+    for bagInfoObject in identifiers:
+        bag = bagInfoObject.belong_to_bag
+        if bag not in unique_identifiers:
+            unique_identifiers.append(bag)
+            data.append(
+                {
+                    'name': bag.name,
+                    'oxum': bag.oxum,
+                    'bagging_date': bag.bagging_date.strftime('%Y-%m-%d')
+                })
```
ldko, line 690: Unlike in `externalIdentifierSearch`, we don't need that much data for each item returned. What do you think about changing the query, so `distinct` is useful? I am thinking something like:
```
    identifiers = (External_Identifier.objects
                                      .select_related('belong_to_bag')
                                      .filter(value=ark)
                                      .values_list('belong_to_bag__name',
                                                   'belong_to_bag__size',
                                                   'belong_to_bag__files',
                                                   'belong_to_bag__bagging_date')
                                      .distinct())

    data = [
        {
            'name': exid[0],
            'oxum': '%s.%s' % (exid[1], exid[2]),
            'bagging_date': exid[3].strftime('%Y-%m-%d')
        }
        for exid in identifiers
    ]
```

ldko (commented):
What you have is working for me. However, I have a couple suggestions to consider.

coda/coda_mdstore/views.py
madhulika95b: oh..yeah..sure. I should have thought of it.

coda/coda_mdstore/views.py
```diff
@@ -670,21 +670,44 @@ def externalIdentifierSearch(request, identifier=None):
 def externalIdentifierSearchJSON(request):
     """External_Identifier search formatted in JSON."""
     ark = request.GET.get('ark', '')
+    showAll = request.GET.get('showAll', False)
```
ldko, line 673: Perhaps the default should be a `str` such as `'false'`, since if there is a value, it would be a `str`.

coda/coda_mdstore/views.py
```diff
@@ -670,21 +670,44 @@ def externalIdentifierSearch(request, identifier=None):
...
     if ('ark:/%d' % settings.ARK_NAAN) not in ark:
         ark = 'ark:/%d/%s' % (settings.ARK_NAAN, ark)
+    data = []
+    if not showAll:
```
ldko, line 677: Currently with this condition, it works as expected if `showAll` is not in the request at all, but if the request has parameter `showAll=False`, `showAll` is set to 'False' which is a non-empty string which evaluates to True. Thus it registers as the same as `showAll=True` and similarly is the case if `showAll` is used with any value.

coda/coda_mdstore/views.py
```diff
@@ -670,21 +670,44 @@ def externalIdentifierSearch(request, identifier=None):
...
+                                          .order_by('-belong_to_bag__bagging_date'))
+
+        data = [
+            {
+                'name': exid[0],
+                'oxum': '%s.%s' % (exid[1], exid[2]),
+                'bagging_date': exid[3].strftime('%Y-%m-%d')
+            }
+            for exid in identifiers
+        ]
```
ldko, line 710: I don't think we need two wholly separate cases here. What if we did:
```
    identifiers = (External_Identifier.objects
                                      .select_related('belong_to_bag')
                                      .filter(value=ark)
                                      .order_by('-belong_to_bag__bagging_date')
                                      .values_list('belong_to_bag__name',
                                                   'belong_to_bag__size',
                                                   'belong_to_bag__files',
                                                   'belong_to_bag__bagging_date')
                                      .distinct())
    if identifiers:
        if showAll.lower() != 'true':
            identifiers = identifiers[:1]
        data = [
            {
                'name': exid[0],
                'oxum': '%s.%s' % (exid[1], exid[2]),
                'bagging_date': exid[3].strftime('%Y-%m-%d')
            }
            for exid in identifiers
        ]
```

coda/coda_mdstore/views.py
madhulika95b: Yeah..I'll have it changed.

ldko (approved):
This looks good to me. Thank you, Madhulika! :+1:

Issue #121, opened 2019-05-24
Add active status for Node


ldko:
It would be helpful for storage maintenance/management if we had a `status` field on the `Node` model that would indicate if the node is active/inactive. We could then use it to determine if we want to disable replication to a storage node etc.

madhulika95b:
Do we want to display the `status` field in `node_status.html` page along with rest of the fields?

vphill:
Here is a rough use case that I'm thinking of. 

> In the situation where we need to replace the storage where node-1 is sitting.  We could mark that node as inactive so that we don’t continue adding to it.  We could move it to a new storage location, update the node in the database and then mark it as active again

In this case I think it makes sense to add a column to the node_status.html to show if the node is active/inactive 

For the XML list of active nodes we should not include inactive nodes so that tools that make use of this won't have to change.

PR #152, opened 2020-09-16
Update node view

### LINKED ISSUE #144: Add sorting to storage node stats
The storage node status page, controlled by [this view](https://github.com/unt-libraries/coda/blob/master/coda/coda_mdstore/views.py#L730) should allow for sorting on at least the names. This will also require changes to the [template](https://github.com/unt-libraries/coda/blob/master/coda/templates/mdstore/node_status.html).

madhulika95b:
@ldko @somexpert 

This address #121 by adding `status` field to `Node` model.
It also closes #144 

After cloning the project, run migrations, create a super user and start the app.
You can add nodes from Django admin site.

After adding the nodes, you can see them along with added status field in node.html, node_status.html pages (in "node/" & "node/(<identifier>" urls). 
You can also see that inactive nodes no longer appear in XML list of nodes (in "APP/node/" url).

You can use some dummy data like below to create a Node object.
node Name: coda-123
node url: http://dummy/coda-123/
node path: http://dummy/coda-123/


coda/templates/mdstore/node.html
```diff
@@ -31,6 +31,10 @@ <h3 class="main">
             <th>Full</th>
             <td> <i class="icon-dashboard"></i> {{ filled }}</td>
         </tr>
+        <tr>
+            <th>Status</th>
+            <td> <i class="icon-dashboard"></i> {{ node.get_status_display }}</td>
```
somexpert, line 36: For this and the other modified template, I think we should change the icon used to be something like maybe a power button or something along those lines.

coda/templates/mdstore/node_status.html
```diff
@@ -38,6 +41,7 @@ <h3>
 				<td> <i class="icon-hdd"></i> {{ status.node.node_size|filesizeformat }} </td>
 				<td> <i class="icon-hdd"></i> {{ status.available|filesizeformat }} </td>
 				<td> <i class="icon-dashboard"></i> {{ status.filled }}% </td>
+                                <td> <i class="icon-calendar"></i> {{ status.node.get_status_display }} </td>
```
somexpert, line 44: Looks like alignment is off here

somexpert (commented):
Looks like the changes are working fine, just have a couple minor comments.

somexpert:
Oh and @ldko, lest you encounter the same confusion I did, here's how `get_status_display` works. It's a built-in feature for choice fields. https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.get_FOO_display

coda/coda_mdstore/models.py
```diff
@@ -70,6 +70,10 @@ class Node(models.Model):
     This model defines a storage node for the mdstore
     """
 
+    STATUS_CHOICES = [
+        ('1', 'Active'),
+        ('2', 'Inactive'),
```
ldko, line 75: Since we are using digits here, anyone else think it would make sense to use '0' for inactive an '1' for active?

somexpert: The thought _did_ occur to me. That or even using just a boolean field: https://docs.djangoproject.com/en/2.2/ref/models/fields/#booleanfield. But I also don't think it's an issue the way it currently is, so I'm fine either way

madhulika95b: I thought of using Boolean Field initially, but then I thought we may want to see 'active/inactive' as status in html pages rather than having a boolean value. I don't mind changing the choices to use '0' and '1'.

coda/templates/mdstore/node_status.html
madhulika95b: I'll fix it.

ldko: The issue is that this line is using spaces for the leading whitespace, while those around it are using tabs. We should be consistent.

madhulika95b: Ah, I see. I'll fix it. Thanks!

coda/coda_mdstore/models.py
somexpert: You got the 0 and 1, but we need to swap which one means active. The intent was for 0 to be _inactive_ and 1 to be _active_.

coda/templates/mdstore/node.html
somexpert: I personally like the move to the green/red highlights, but we should see if others like it or would prefer to stick with icons for consistency.

somexpert: thoughts @ldko?

madhulika95b: We have similar setup for Premis events. I just used the same idea. :)

ldko: I think the red and green text background is fine.

somexpert (approved):
LGTM :thumbsup:

coda/templates/mdstore/node.html
```diff
@@ -35,5 +35,9 @@ <h3 class="main">
             <th>Last Updated</th>
             <td> <i class="icon-calendar"></i> {{ node.last_checked }}</td>
         </tr>
+        <tr>
+	    <th>Status</th>
+	    <td> <span title="{{ node.status }}" class="disabled btn btn-primary btn-{% if node.status == "0" %}success{% else %}danger{% endif %}">{{ node.get_status_display }}</span></td>
+        </tr>
```
ldko, line 41: I think we should swap the colors to where `status` 0 should be danger, and 1 should be success here and for the node_status.html.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -780,13 +780,14 @@ def test_get_request_without_identifier(self, rf):
...
 
     def test_get_request_without_identifier_response_content(self, rf):
-        NodeFactory.create_batch(10)
+        NodeFactory.create_batch(7, status="1")
+        NodeFactory.create_batch(5, status="0")
         request = rf.get('/', HTTP_HOST='example.com')
         response = views.app_node(request)
         tree = objectify.fromstring(response.content)
 
         # Check that all of the Nodes are listed in the feed.
```
ldko, line 789: We could update this comment to say "all of the active Nodes".

ldko (commented):
Perhaps we should add `status` to the the node XML in the feed (though it should always be "Active" if it is in the feed) and for individual node identifier (APP/node/coda-002) if it exists. I think that would happen in [codalib](https://github.com/unt-libraries/codalib/blob/3c8e30097de74a300f7c503802117361a98e5261/codalib/bagatom.py#L200). What do you think?

Also, perhaps we could add the `status` to the [admin view](https://github.com/unt-libraries/coda/blob/f613b763769f4f8176f05547e7eb6434dde5fd69/coda/coda_mdstore/admin.py#L21).

madhulika95b:
> Perhaps we should add `status` to the the node XML in the feed (though it should always be "Active" if it is in the feed) and for individual node identifier (APP/node/coda-002) if it exists. I think that would happen in [codalib](https://github.com/unt-libraries/codalib/blob/3c8e30097de74a300f7c503802117361a98e5261/codalib/bagatom.py#L200). What do you think?
> 
> Also, perhaps we could add the `status` to the [admin view](https://github.com/unt-libraries/coda/blob/f613b763769f4f8176f05547e7eb6434dde5fd69/coda/coda_mdstore/admin.py#L21).

I'll add the `status` to xml feed in codalib and open a PR and after merging it we can have this closed. Does that sound ok?

ldko (approved):
Looks good to me :+1: . Thank you, Madhulika!

PR #153, opened 2020-09-30
Update settings

### LINKED ISSUE #132: Update secrets.json.template
I think we should include all the settings that are required to get a production instance running. As it is right now, we would get errors if we based a production deployment off this, since the production settings would attempt to retrieve settings that won't be there in the secrets file, such as the SITE_ID.

madhulika95b:
@ldko @somexpert 

This PR adds some settings to `secrets.json.template` and closes #132

### DIFF
--- a/docker_secrets.json
+++ b/docker_secrets.json
@@ -9,5 +9,5 @@
   "ALLOWED_HOSTS": [],
   "SITE_ID": "",
   "ADMINS": "",
-  "STATIC_ROOT": "",
+  "STATIC_ROOT": ""
 }
--- a/docker_secrets.json
+++ b/docker_secrets.json
@@ -5,5 +5,9 @@
   "DB_USER": "",
   "DB_HOST": "",
   "DB_PORT": "",
-  "DB_PASSWORD": ""
+  "DB_PASSWORD": "",
+  "ALLOWED_HOSTS": [],
+  "SITE_ID": "",
+  "ADMINS": "",
+  "STATIC_ROOT": "",
 }
--- a/secrets.json.template
+++ b/secrets.json.template
@@ -9,5 +9,5 @@
   "ALLOWED_HOSTS": [],
   "SITE_ID": "",
   "ADMINS": "",
-  "STATIC_ROOT": "",
+  "STATIC_ROOT": ""
 }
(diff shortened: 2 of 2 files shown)


docker_secrets.json
```diff
@@ -5,5 +5,9 @@
   "DB_USER": "",
   "DB_HOST": "",
   "DB_PORT": "",
-  "DB_PASSWORD": ""
+  "DB_PASSWORD": "",
+  "ALLOWED_HOSTS": [],
+  "SITE_ID": "",
+  "ADMINS": "",
```
ldko, line 11: ADMINS should probably be an empty list.

docker_secrets.json
```diff
@@ -5,5 +5,9 @@
   "DB_USER": "",
   "DB_HOST": "",
   "DB_PORT": "",
-  "DB_PASSWORD": ""
+  "DB_PASSWORD": "",
+  "ALLOWED_HOSTS": [],
+  "SITE_ID": "",
```
ldko, line 10: Usually SITE_ID is an int, not sure if we should change it from empty string.

docker_secrets.json
```diff
@@ -5,5 +5,9 @@
...
+  "ADMINS": "",
+  "STATIC_ROOT": ""
 }
```
ldko, line 13: Do we need to include `STATIC_URL` also?

madhulika95b: Oh, yes! I thought `STATIC_URL` in `coda/config/settings/base.py` should be fine, but looking at it again, as we are using `get_secret` to get the value in `production.py`, we'll require it in `secrets.json`. I'll have it added.

docker_secrets.json
madhulika95b: Yeah...I'll have it changed.

ldko: I think that would be ok. What do you think @somexpert ?

somexpert: I think it's fine. We're certainly not advocating that people deploy this into a production setting using that Docker setup.

ldko (approved):
Looks fine to me. Thank you, Madhulika! :+1:

PR #154, opened 2020-10-06
Add file links

### LINKED ISSUE #118: Add view of data file links
@vphill would like a view that serves an HTML file with links to all of the data files a user could download for a bag (i.e. all files available for a bag that can be downloaded using `coda_mdstore.views.bagProxy`).

madhulika95b:
@ldko @somexpert 
This PR adds a template to display all downloadable bag files. This closes #118

### DIFF
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -688,10 +688,11 @@ class TestBagURLListView:
             b'']
         bag = FullBagFactory.create()
         request = rf.get('/')
-        response = views.bagURLList(request, bag.name, bagfiles='bagfiles')
-        assert b'<a' in response.content
-        assert b'https://coda/data/file02.txt</a>' in response.content
-        assert b'https://coda/data/file01.txt</a>' in response.content
+        response = views.bagURLList(request, bag.name, html=True)
+        assert (b'<a href="https://coda/data/file02.txt"> '
+                b'https://coda/data/file02.txt</a><br>') in response.content
+        assert (b'<a href="https://coda/data/file01.txt"> '
+                b'https://coda/data/file01.txt</a><br>') in response.content
 
 
 class TestBagFullTextSearchHTMLView:
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -622,7 +622,7 @@ class TestBagURLListView:
 
         self.getFileHandle = mock.Mock(return_value=file_handle)
         monkeypatch.setattr(
-            'coda_mdstore.views.getFileHandle', self.getFileHandle)
+            'coda_mdstore.presentation.getFileHandle', self.getFileHandle)
 
     @pytest.mark.xfail(reason='The response content does not match other '
                               'responses.')
@@ -668,7 +668,7 @@ class TestBagURLListView:
         bag = FullBagFactory.create()
         request = rf.get('/')
         # Mock file names found at the bag's root level.
-        with mock.patch('coda_mdstore.views.getFileList',
+        with mock.patch('coda_mdstore.presentation.getFileList',
                         return_value=['bagit.txt', 'bag-info.txt']):
             response = views.bagURLList(request, bag.name)
 
--- a/coda/coda_mdstore/tests/test_views.py
+++ b/coda/coda_mdstore/tests/test_views.py
@@ -622,7 +622,7 @@ class TestBagURLListView:
 
         self.getFileHandle = mock.Mock(return_value=file_handle)
         monkeypatch.setattr(
-            'coda_mdstore.presentation.getFileHandle', self.getFileHandle)
+            'coda_mdstore.views.getFileHandle', self.getFileHandle)
 
     @pytest.mark.xfail(reason='The response content does not match other '
                               'responses.')
@@ -668,7 +668,7 @@ class TestBagURLListView:
         bag = FullBagFactory.create()
         request = rf.get('/')
         # Mock file names found at the bag's root level.
-        with mock.patch('coda_mdstore.presentation.getFileList',
+        with mock.patch('coda_mdstore.views.getFileList',
                         return_value=['bagit.txt', 'bag-info.txt']):
             response = views.bagURLList(request, bag.name)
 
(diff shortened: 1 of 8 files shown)


coda/coda_mdstore/views.py
```diff
@@ -532,7 +532,7 @@ def bagHTML(request, identifier):
     )
 
 
-def bagURLList(request, identifier):
+def bagURLList(request, identifier, bagfiles=None):
```
somexpert, line 535: I propose we rename the kwarg something like `rendered` or `html`, and default it to `False` here and pass it as `True` in the new URL.

coda/coda_mdstore/tests/test_views.py
```diff
@@ -678,6 +678,21 @@ def test_response_content_has_topFiles(self, rf):
...
+        self.getFileHandle.return_value.url = 'https://coda/testurl'
+        # Mock what gets read from the manifest file.
+        self.getFileHandle.return_value.readline.side_effect = [
+            b'192e635b17a9c2aea6181f0f87cab05d  data/file01.txt',
+            b'18b7c500ef8bacf7b2151f83d28e7ca1  data/file02.txt',
+            b'']
+        bag = FullBagFactory.create()
+        request = rf.get('/')
+        response = views.bagURLList(request, bag.name, bagfiles='bagfiles')
+        assert b'<a' in response.content
```
somexpert, line 692: I don't really think this assertion by itself adds much. Instead, why don't we add the full <a> tag (opening and closing) to the next two assertions so they each check not only that they URL is present, but also that it's in link form?

coda/templates/mdstore/bag_files_download.html
```diff
@@ -0,0 +1,15 @@
+{% extends "mdstore/base.html" %}
+{% block content %}
+{% load humanize %}
```
somexpert, line 3: I'm not quite sure how to load content into my local instance, so I'm not sure how this page actually looks, nor do I know if there is a desired look for it, but if it's just a straight list of links, I don't think we really need humanize loaded in here.

somexpert (commented):
Just a few comments

coda/templates/mdstore/bag_files_download.html
```diff
@@ -0,0 +1,15 @@
...
+    {%for item in outputText%}
+        <tr>
+            <td>
+                <a href="item" download>{{ item }}</a>
+            </td>
+        </tr>
+    {% endfor %}
+</table>
+</body>
+</html>
```
ldko, line 14: I think we could simplify this by leaving out the `<table>` and also we don't need to include the `</body>` and `</html>` tags, since they will be included by the `endblock`. Also, currently the links are not working. Perhaps we could use:
```suggestion
{% for item in outputText %}
    <a href="{{ item }}">{{ item }}</a><br>
{% endfor %}
```

coda/coda_mdstore/urls.py
```diff
@@ -9,6 +9,10 @@
     re_path(r'^APP/bag/(?P<identifier>.+?)/$', views.app_bag, name='app-bag-detail'),
     re_path(r'^bag/(?P<identifier>.+?)/$', views.bagHTML, name='bag-detail'),
     re_path(r'^bag/(?P<identifier>ark:\/\d+\/.+?).urls$', views.bagURLList, name='bag-urls'),
+    re_path(
+        r'^bag/(?P<identifier>ark:\/\d+\/.+?)/bagfiles$', views.bagURLList,
+        kwargs={'bagfiles': 'bagfiles'}, name='bag-files'
+    ),
```
ldko, line 15: What do you all think if we change the path of this to `/downloads` or `/links`? Also, since his does not have a file extension, should it have a slash on the end?

coda/coda_mdstore/views.py
ldko: I agree with Gio about a Boolean value here. I like `html` as the kwarg.

ldko (changes requested):
In addition to adding this view, I think it would also be helpful to users if we add a button on the main bag html page to this page (next to the URLS and ATOM buttons).

coda/templates/mdstore/bag_files_download.html
madhulika95b: Yeah...I'll have to change that to `<a href="{{ item }}">`. I was thinking we'd want to have downloadable links here. The way we have now along with the above mentioned change is, on clicking the link it loads the web page, but not download those files.

coda/coda_mdstore/urls.py
somexpert: I'm on board with either of those and also the trailing slack

coda/templates/mdstore/bag_search_results.html
```diff
@@ -50,6 +53,9 @@
                     <td class="data" style="vertical-align:middle">
                         <i class="icon-globe"></i> <a href="http://{{ request.META.HTTP_HOST }}/APP/bag/{{ entry.name }}">ATOM</a>
                     </td>
+                    <td class="data" style="vertical-align:middle">
+                        <i class="icon-globe"></i> <a href="http://{{ request.META.HTTP_HOST }}/bag/{{ entry.name }}/links/">links</a>
```
ldko, line 57: What do you think about changing the links for this view and the others to use the `{% url %}` tag?

coda/templates/mdstore/bag_info.html
```diff
@@ -6,6 +6,7 @@
     <a class="btn btn-large btn-primary">{{ bag }}</a>
     <a class="btn btn-large btn-warning" href="{{ request.build_absolute_uri|slice:":-1" }}.urls">URLS</a>
     <a class="btn btn-large btn-danger" href="http://{{ request.META.HTTP_HOST }}/APP/bag/{{ bag }}">ATOM</a>
+    <a class="btn btn-large btn-success" href="http://{{ request.META.HTTP_HOST }}/bag/{{ bag }}/links/">Links</a>
```
ldko, line 9: Would it work to use the `{% url %}` tag for these buttons' hrefs?

coda/templates/mdstore/bag_files_download.html
```diff
@@ -0,0 +1,6 @@
+{% extends "mdstore/base.html" %}
+{% block content %}
+    {%for item in outputText%}
+        <a href="{{ item }}"> {{ item }}</a><br>
```
ldko, line 4: The space before the second ` {{ item }}` seems superfluous.

coda/templates/mdstore/bag_files_download.html
```diff
@@ -0,0 +1,6 @@
+{% extends "mdstore/base.html" %}
+{% block content %}
+    {%for item in links%}
```
ldko, line 3: For consistency and readability, would you add a space after the first `%` and before the second one?

somexpert (approved):
LGTM :thumbsup:

ldko (approved):
Thanks, Madhulika! Looks good!

PR #156, opened 2020-10-13
Fix intermittent status test fail

ldko:
After last merge into master, this test failed (had not failed in CI check during PR) which highlighted that the test intermittently fails if the number of entries created with `status` of 2 was greater than 20 because of pagination breaking entries at 20. This change explicitly sets the number of entries with and without a `status` of '2' and keeps them below the pagination number.

This is ready for review @madhulika95b @somexpert .


somexpert (approved):
LGTM :thumbsup:

PR #157, opened 2020-10-20
Bag download

### LINKED ISSUE #117: Allow downloading of zipped bags
Idea is to allow a user who is viewing a bag to click a button to download a zip file (generated on the fly) of the full bag (not just the bag's data files).

@vphill linked to https://github.com/SpiderOak/ZipStream

madhulika95b:
@ldko @somexpert 
This MR closes #117 using `zipstream` module to stream the downloadable zipped bag file on the fly.

### DIFF
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -92,13 +92,12 @@ def file_chunk_generator(url):
         yield chunk
 
 
-def zip_file_streamer(urls, identifier):
+def zip_file_streamer(urls, identifier, meta_id):
     """
-    Stream downloadable zipped file using zipstream
+    Stream zipped file using zipstream
     """
     with zipstream.ZipFile(mode='w') as zip_obj:
         for url in urls:
-            meta_id = identifier.split('/')[-1]
             filename = '%s/%s' % (meta_id, url.split(meta_id, 1)[-1])
 
             zip_obj.write_iter(filename, file_chunk_generator(url))
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -4,7 +4,6 @@ import urllib.request
 import urllib.parse
 import requests
 import zipstream
-from zipfile import ZIP_DEFLATED
 
 from bs4 import BeautifulSoup as BSoup
 from codalib import APP_AUTHOR
@@ -93,19 +92,20 @@ def file_chunk_generator(url):
         yield chunk
 
 
-def zip_file_streamer(urls):
+def zip_file_streamer(urls, identifier):
     """
     Stream downloadable zipped file using zipstream
     """
-    zip_obj = zipstream.ZipFile(mode='w', compression=ZIP_DEFLATED)
-    for url in urls:
-        filename = os.path.basename(url)
-        zip_obj.write_iter(filename, file_chunk_generator(url))
+    with zipstream.ZipFile(mode='w') as zip_obj:
+        for url in urls:
+            meta_id = identifier.split('/')[-1]
+            filename = '%s/%s' % (meta_id, url.split(meta_id, 1)[-1])
 
-    # Each call will iterate the generator one at a time until all files are completed.
-    for chunk in zip_obj:
-        yield chunk
-    zip_obj.close()
+            zip_obj.write_iter(filename, file_chunk_generator(url))
+
+        # Each call will iterate the generator one at a time until all files are completed.
+        for chunk in zip_obj:
+            yield chunk
 
 
 def bagSearch(bagString):
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -2,6 +2,9 @@ import os
 import re
 import urllib.request
 import urllib.parse
+import requests
+import zipstream
+from zipfile import ZIP_DEFLATED
 
 from bs4 import BeautifulSoup as BSoup
 from codalib import APP_AUTHOR
@@ -77,6 +80,33 @@ def getFileHandle(codaId, codaPath):
     raise Exception(
         "Unable to get handle for id %s at path %s" % (codaId, codaPath)
     )
+
+
+def file_chunk_generator(url):
+    """
+    Download a file and stream it
+    """
+    r = requests.get(url, stream=True)
+    if r.status_code != 200:
+        return
+    for chunk in r.iter_content(1024):
+        yield chunk
+
+
+def zip_file_streamer(urls):
+    """
+    Stream zip file
+    """
+    zip_obj = zipstream.ZipFile(mode='w', compression=ZIP_DEFLATED)
+    for url in urls:
+        filename = os.path.basename(url)
+        zip_obj.write_iter(filename, file_chunk_generator(url))
+
+    # Each call will iterate the generator we wrote for each file
+    # one at a time until all files are completed.
+    for chunk in zip_obj:
+        yield chunk
+    zip_obj.close()
 
 
 def bagSearch(bagString):
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -92,7 +92,7 @@ def file_chunk_generator(url):
         yield chunk
 
 
-def zip_file_streamer(urls, identifier, meta_id):
+def zip_file_streamer(urls, meta_id):
     """
     Stream zipped file using zipstream
     """
--- a/coda/coda_mdstore/presentation.py
+++ b/coda/coda_mdstore/presentation.py
@@ -95,15 +95,14 @@ def file_chunk_generator(url):
 
 def zip_file_streamer(urls):
     """
-    Stream zip file
+    Stream downloadable zipped file using zipstream
     """
     zip_obj = zipstream.ZipFile(mode='w', compression=ZIP_DEFLATED)
     for url in urls:
         filename = os.path.basename(url)
         zip_obj.write_iter(filename, file_chunk_generator(url))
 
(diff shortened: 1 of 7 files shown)


coda/coda_mdstore/views.py
```diff
@@ -589,6 +589,13 @@ def bagURLList(request, identifier, html=False):
         return render(request, 'mdstore/bag_files_download.html',
                       {'links': sorted(transList)})
 
+    if download:
+        zip_filename = 'bag-' + identifier.replace('/', '_') + '.zip'
```
ldko, line 593: I'm not particularly fond of this file naming convention, though I suppose it doesn't really matter. What do people (@vphill @madhulika95b @somexpert ) think about just having the coda meta_id like coda101.zip or dropping the 'bag-' prefix?

coda/coda_mdstore/views.py
```diff
@@ -589,6 +589,13 @@ def bagURLList(request, identifier, html=False):
...
+        response = StreamingHttpResponse(zip_file_streamer(transList),
+                                         content_type='application/x-zip-compressed')
```
ldko, line 595: Do you know what the purpose of using 'applicaiton/x-zip-compressed' is over 'application/zip'?

coda/coda_mdstore/urls.py
```diff
@@ -11,6 +11,10 @@
         r'^bag/(?P<identifier>ark:\/\d+\/.+?)/links/$', views.bagURLList,
         kwargs={'html': True}, name='bag-links'
     ),
+    re_path(
+        r'^bag/(?P<identifier>ark:\/\d+\/.+?)/download/$', views.bagURLList,
+        kwargs={'download': True}, name='bag-downloads'
```
ldko, line 16: How about singular 'bag-download' for the `name` since it is downloading a single bag?

ldko (changes requested):

coda/coda_mdstore/views.py
vphill: I would prefer the bag name to be the `metaid + '.zip'` like `coda101.zip`

coda/coda_mdstore/urls.py
vphill: it might make sense to have the path for these downloadable zip files something like `http://coda.library.unt.edu/bag/ark:/67531/coda5kd3.zip` 

I don't think we are 100% consistent in the URL patterns in coda and that is probably beyond the scope here.  For me it would mirror the `.urls` files we return.

coda/coda_mdstore/views.py
madhulika95b: I'm not sure what the difference is and which is preferred one. I'll look up about it.

coda/coda_mdstore/views.py
```diff
@@ -532,7 +532,7 @@ def bagHTML(request, identifier):
     )
 
 
-def bagURLList(request, identifier, html=False):
+def bagURLList(request, identifier, html=False, download=False):
     """
```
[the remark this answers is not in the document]
madhulika95b: yes! I can move this `transList` generation to `presentation.py` and use it in views and separate them.

coda/coda_mdstore/presentation.py
```diff
@@ -79,6 +82,32 @@ def getFileHandle(codaId, codaPath):
...
+        return
+    for chunk in r.iter_content(1024):
+        yield chunk
+
+
+def zip_file_streamer(urls):
+    """
+    Stream downloadable zipped file using zipstream
+    """
+    zip_obj = zipstream.ZipFile(mode='w', compression=ZIP_DEFLATED)
```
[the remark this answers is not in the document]
madhulika95b: sure, I'll try and see if it will work.

coda/coda_mdstore/presentation.py
```diff
@@ -79,6 +82,32 @@ def getFileHandle(codaId, codaPath):
     )
 
 
+def file_chunk_generator(url):
+    """
+    Download a file and stream it
+    """
+    r = requests.get(url, stream=True)
+    if r.status_code != 200:
+        return
```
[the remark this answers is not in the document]
madhulika95b: Do you suggest to halt downloading and raise an exception even if one (or few) of the files is/are invalid? or do we want to continue with the download with available files and have an exception raised when all the file urls are invalid to avoid empty bag downloade?

coda/coda_mdstore/views.py
madhulika95b: `application/zip` is a standard Mime type for zip files. `applicaiton/x-zip-compressed` is a variation of it which is used to denote the encoding necessary for the ZIP file format.

coda/coda_mdstore/presentation.py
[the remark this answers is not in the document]
ldko: I have been looking at this and am not sure of the best path forward. At the moment I am leaning toward leaving it as you have it and leaving it up to the user to verify the bag is valid...not sure how to otherwise handle an exception while streaming the chunks of files...

coda/coda_mdstore/views.py
vphill: Is there an assumption that `application/x-zip-compressed` actually makes use of compression in the file (I don't think we are here right?) vs `application/zip`?

madhulika95b: Honestly I'm not sure if I should be using `application/x-zip-compressed` or `application/zip`.

But, we are using `ZIP_DEFLATED` compression when writing files to `ZipFile`.
`zipstream.ZipFile(mode='w', compression=ZIP_DEFLATED)`

vphill: I think that we should give `compression= ZIP_STORED` (might be the default value) a try instead of `ZIP_DEFLATED`.  If I understand this correctly, it would store the file as-is in the zip file without additional compression, much like a tar file.  If that seems to work then I think `application/zip` would be just fine for the content_type. 

Another note,  do we need to set `allowZip64=True` (https://docs.python.org/3.8/library/zipfile.html#zipfile-objects) because we will be creating zips bigger than 4Gb sometimes?

ldko: `allowZip64=True` is the default since Python 3.4.

ldko (approved):
Looks good to me and worked on an 89GB bag :+1:  Thanks, Madhulika!

coda/coda_mdstore/presentation.py
```diff
@@ -79,6 +81,33 @@ def getFileHandle(codaId, codaPath):
...
+    for chunk in r.iter_content(1024):
+        yield chunk
+
+
+def zip_file_streamer(urls, identifier):
+    """
+    Stream downloadable zipped file using zipstream
```
somexpert, line 97: We can remove "downloadable", I think in this context it's redundant with "Stream".

coda/coda_mdstore/presentation.py
```diff
@@ -79,6 +81,33 @@ def getFileHandle(codaId, codaPath):
...
+    """
+    with zipstream.ZipFile(mode='w') as zip_obj:
+        for url in urls:
+            meta_id = identifier.split('/')[-1]
```
somexpert, line 101: Let's pull this out of the loop, we don't need to be repeating this processing for each url, since the identifier isn't changing.

coda/coda_mdstore/presentation.py
```diff
@@ -79,6 +81,33 @@ def getFileHandle(codaId, codaPath):
...
+            filename = '%s/%s' % (meta_id, url.split(meta_id, 1)[-1])
```
somexpert, line 102: I _kind_ of wonder if this would be better suited for the `re` module, but I'm not sure it's warranted. But that's just a thought, take it or leave it :smile:

coda/coda_mdstore/urls.py
```diff
@@ -11,6 +11,10 @@
         r'^bag/(?P<identifier>ark:\/\d+\/.+?)/links/$', views.bagURLList,
         kwargs={'html': True}, name='bag-links'
     ),
+    re_path(
+        r'^bag/(?P<identifier>ark:\/\d+\/.+?).zip$', views.bagURLList,
```
somexpert, line 15: Is this regex correct? I feel like the `.+?` part is odd. Should it just be `.*`?

somexpert:
Also, should we get some test coverage for the two new functions in the presentation file?

coda/coda_mdstore/urls.py
ldko: or just `.+` without the `?` because you would expect it to have atleast one char there.

coda/coda_mdstore/presentation.py
ldko: You could also pass the metaid part only to the `zip_file_streamer` function rather than the full identifier.

madhulika95b:
I'm planning to move the generation of transList files in `bagURLList` view into a separate function in a separate PR. What do you think about addressing these tests along with tests for the new function in that PR?

ldko:
If the presentation functions added in this PR are presumably not affected by the `bagURLList` view being split up, I think it would make more sense to add the presentation tests to this PR or to open an issue and do a separate PR.

madhulika95b:
ok, I'll open an issue and write tests for these in a separate PR.

coda/coda_mdstore/presentation.py
somexpert: Like @ldko was saying, since we now pass in `meta_id`, we can just get rid of the `identifier` arg.

madhulika95b: oh...yeah. Will get rid of it.

somexpert (approved):
Looks good to me! :thumbsup:

PR #162, opened 2021-01-06
Update to mariadb

madhulika95b:
@ldko @somexpert 

This PR upgrades the db from 'mysql' to 'mariab'

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,7 @@ ENV PYTHONPATH /app:/app/coda
 RUN mkdir /app
 WORKDIR /app
 
-RUN apt-get update -qq && \
-    apt-get install -y  mariadb-server mariadb-client libmariadbclient-dev libssl-dev netcat
+RUN apt-get update -qq && apt-get install -y mysql-client netcat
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -11,7 +11,8 @@ ENV PYTHONPATH /app:/app/coda
 RUN mkdir /app
 WORKDIR /app
 
-RUN apt-get update -qq && apt-get install -y mysql-client netcat
+RUN apt-get update -qq && \
+    apt-get install -y  mariadb-server mariadb-client libmariadbclient-dev libssl-dev netcat
```
somexpert, line 15: So did you try without changing this line? I don't think we needed to change the client, mysql-client should work just fine connecting to mariadb.

madhulika95b: you are right! `mysql-client` works just fine. I will revert it to what it was.

somexpert (approved):
Looks good to me! :thumbsup:

ldko (approved):
works for me! :+1:  Thank you, Madhulika!

PR #167, opened 2021-05-04
Update about page

madhulika95b:
@ldko  @somexpert 

This PR updates the url examples in about page. The fragment '#failure' and '#success' is getting stripped off by the browser and never sent to server by Http Request and there '#' should be escaped in the url using '%23'.

### DIFF
--- a/coda/templates/mdstore/about.html
+++ b/coda/templates/mdstore/about.html
@@ -15,7 +15,7 @@
     <!--BEGIN CODE MARKUP-->
     <div id="highlight-python"><pre>
     # to get only the events which have failed, use the 'outcome' url argument
-    $ curl {{ request.META.HTTP_HOST }}/event/search.json?event_outcome=http%3A%2F%2Fpurl.org%2Fnet%2Funtl%2Fvocabularies%2FeventOutcomes%2F%23failure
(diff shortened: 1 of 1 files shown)


coda/templates/mdstore/about.html
```diff
@@ -35,7 +35,8 @@ <h3>Search API</h3>
     ]
 
     # to get the events which failed after a certain date, append a 'start_date' argument
-    $ curl {{ request.META.HTTP_HOST }}/event/search.json?outcome=http://purl.org/net/untl/vocabularies/eventOutcomes#failure&start_date=05/13/2013
+    $ curl {{ request.META.HTTP_HOST }}/event/search.json?event_outcome=http://purl.org/net/untl/vocabularies/eventOutcomes/%23failure&start_date=05/13/2013
```
ldko, line 38: Trying this one with `curl`, I needed to put the url into quotes because of the `&` before `start_date` which otherwise is interpreted to run the command in the background.

ldko (approved):
Works for me. Thank you, Madhulika!

ldko:
Hmm, not sure why a test failed...seems unrelated to these changes.

somexpert (approved):
This looks good to me and worked in testing (and the tests all passed for me in Docker). :thumbsup:

PR #172, opened 2023-02-23
Upgrade to Django4

gracieflores:
One change in the `requirements.txt` file is temporary, and that is the installing of `django-premis-event-service` django4 branch. One other change that I wasn't sure about was the `wsgi.py` file (`SITE_PACKAGES` change).

@ldko @somexpert This is ready for review.


ldko (approved):
I ran this in Docker, POSTed a couple bags and queue entry. It works for me.

somexpert (approved):
Changes look good, including the wsgi changes, and the tests ran for me in podman. :thumbsup:

gracieflores:
@ldko @somexpert Since I have made changes to the settings, would you please review those changes. The changes were successful when I ran the podman instance and the tests passed.

ldko:
@gracieflores do the changes to the settings still have the Django Debug Toolbar showing on the local deployment?

coda/config/settings.py
```diff
@@ -47,7 +20,13 @@ def get_secret(setting, secrets=secrets):
         raise ImproperlyConfigured(error_msg)
 
 
-DEBUG = True
+DEBUG = get_secret('SECRET_KEY')
```
ldko, line 23:
```suggestion
DEBUG = get_secret('DEBUG')
```

ldko:
We should also update secrets.json.template if new secrets are required.

somexpert:
Changes look good to me, and the app and tests ran fine locally :thumbsup:

ldko (approved):
Thanks, Gracie! :+1:

PR #175, opened 2023-05-26
Add data-upload-max setting

gracieflores:
This is a small change to allow for deleting a larger amount of queue entries (than the default allows) in the admin.


gracieflores:
@ldko @somexpert This is ready for review.

somexpert (approved):
At first I was wondering where we were gonna actually implement this setting, but I see that's it's a Django built-in setting. So it looks good to me! :thumbsup:

ldko (approved):
Looks good to me :+1: . Thank you, Gracie!

PR #177, opened 2024-03-26
Adjust URLs to make coda work on HTTP or HTTPS.

somexpert:
This adjusts things so that we are making use of the correct scheme for our internal URLs whether we're on HTTP or HTTPS.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,5 +16,6 @@ django-debug-toolbar~=3.8.0
 zipstream>=1.1.4
 
 git+https://github.com/unt-libraries/codalib
-git+https://github.com/unt-libraries/django-premis-event-service.git@master
+#TODO: Revert branch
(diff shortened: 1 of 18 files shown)


requirements.txt
```diff
@@ -16,5 +16,6 @@ django-debug-toolbar~=3.8.0
 zipstream>=1.1.4
 
 git+https://github.com/unt-libraries/codalib
-git+https://github.com/unt-libraries/django-premis-event-service.git@master
+#TODO: Revert branch
+git+https://github.com/unt-libraries/django-premis-event-service.git@https-compatible
```
ldko, line 20: Once this is reverted I think it will be ok to merge.

ldko (commented):
This looks good to me. Just need to revert the requirements branch for django-premis-event-service.

PR #178, opened 2024-04-05
Switch jss/css URLs to https

ldko:
@somexpert this updates the js and css URLs to prevent mixed protocol errors.


somexpert (approved):
Looks good to me! :thumbsup:

PR #179, opened 2024-04-05
Add SECURE_PROXY_SSL_HEADER setting

ldko:
@somexpert this sets us up to use the SECURE_PROXY_SSL_HEADER setting, so https will work behind a load balancer.


somexpert (approved):
This looks good to me! :thumbsup:

PR #184, opened 2025-07-07
Favor unverified validation; update `docker compose` usage

ldko:
@clarktr1 @somexpert this is ready for review. The main change of substance here is that when determining what item should be selected for validation next, bags with a `last_verified_status` of `Unverified` will be preferred over random bags that aren't prioritized and have previously been validated at some point. Also included is updating the `docker compose` usage to remove the outdated `docker-compose` command.

### DIFF
--- a/coda/coda_validate/tests/test_feeds.py
+++ b/coda/coda_validate/tests/test_feeds.py
@@ -40,11 +40,20 @@ class TestAtomNextNewsFeed:
         feed.items('')
         assert 'Item was chosen because it is the oldest' in feed.reason
 
+    def test_items_chooses_unverified_validate(self):
+        factories.ValidateFactory.create_batch(1, last_verified_status='Unverified', priority=0)
+        factories.ValidateFactory.create_batch(19, last_verified_status='Passed', priority=0)
+
+        feed = views.AtomNextNewsFeed()
+        feed.items('')
+        assert 'Item was chosen because it is Unverified' in feed.reason
+
     def test_items_chooses_a_random_validate(self):
         # Make the last_verified time twice less than the validation period time
         # to ensure that we always randomly select a Validate object.
         last_verified = datetime.now() - (settings.VALIDATION_PERIOD * 2)
-        factories.ValidateFactory.create_batch(30, priority=0, last_verified=last_verified)
+        factories.ValidateFactory.create_batch(30, priority=0, last_verified=last_verified,
+                                               last_verified_status='Passed')
 
         feed = views.AtomNextNewsFeed()
         feed.items('')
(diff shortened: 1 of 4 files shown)


coda/coda_validate/tests/test_feeds.py
```diff
@@ -40,19 +40,32 @@ def test_items_chooses_oldest_validate(self):
...
+    def test_items_chooses_unverified_validate(self):
+        oldest = factories.ValidateFactory.create_batch(1, last_verified_status='Unverified',
+                                                        priority=0)
+        factories.ValidateFactory.create_batch(18, last_verified_status='Passed', priority=0)
+        factories.ValidateFactory.create_batch(1, last_verified_status='Unverified', priority=0)
+
+        feed = views.AtomNextNewsFeed()
+        feed_queryset = feed.items('')
+        assert oldest[0].identifier == feed_queryset[0].identifier
+        assert 'Item was chosen because it is the oldest Unverified' in feed.reason
```
somexpert, line 53: I think it might be nice to check that there are _two_ items in the feed, as well as verifying that they both have that same reason, just to show that the Unverified items have priority regardless of creation date.

coda/coda_validate/views.py
```diff
@@ -63,31 +63,37 @@ def items(self, obj):
...
+                # http://elpenia.wordpress.com/2010/05/11/getting-random-objects-from-a-queryset-in-django/
+                now = datetime.datetime.now()
+                v = validations.filter(
+                    last_verified__lte=now - settings.VALIDATION_PERIOD
+                )
+                if v.exists():
+                    random_slice = int(random.random() * v.count())
+                    v = v[random_slice:]
+                    reason += 'Item was randomly selected and within the \
+past year because there is no prioritized record.'
```
clarktr1, line 92: How come bags need to get verified randomly instead of giving prio to the first added (similar to a queue)?

ldko: That code was already the logic here, I just shifted the code in the conditional structure for the new condition. It could work to select the items due for a validation (last validation is outside the decided upon validation period) then order it by oldest last verified timestamp, but for some reason it was initially decided this should be a random selection of anything needing its next validation. Here is the overall order of selecting things for validation (though I could see some of it being done differently like you hint):
1. Validate anything that has been actively "prioritized" (i.e. someone has sent a request to the API to change priority status of a bag to 1)
2. Validate anything that has never been validated, based on oldest added that has "Unverified" status (items are ingested into the system with "Unverified" validation status)
3. Randomly select anything that has previously been validated but is due to be validated again because it hasn't been validated in the past year (or in whatever time period is set in the settings)
4. Pick the item that has been validated previously but was validated longest ago (though it's not actually due for another validation according to the time in the settings)

coda/coda_validate/tests/test_feeds.py
ldko: The "next" feed is only meant to contain 1 item, the item that should next have validation run on it, so it ends with  `return v[:1]`. I updated the test to show that is the case.

clarktr1 (approved):
Looks okay to me!

coda/coda_validate/tests/test_feeds.py
somexpert: ah, thanks for the clarification!

somexpert (approved):
Looks good to me! :thumbsup:

